PR #30, opened 2024-03-11
fix: hugging-cli token retrieval issue

rickstaa:
This pull request ensures that the user's authentication token from 'huggingface-cli' is utilized for downloading models when the `HF_TOKEN` environment variable is unset.

Not a necessity just something I use on my own machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ input
 output
 models
 .vscode
+aiModels.json
--- a/runner/dl_checkpoints.sh
+++ b/runner/dl_checkpoints.sh
@@ -2,13 +2,24 @@
 
 mkdir -p models
 
+# Use HF_TOKEN if set, otherwise use huggingface-cli's login.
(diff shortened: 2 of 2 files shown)


runner/README.md
```diff
@@ -35,6 +35,9 @@ pip install "huggingface_hub[cli]"
 ./dl-checkpoints.sh
 ```
 
+> [NOTE]\
+> The `dl-checkpoints.sh` script includes models that are restricted by token gating. By default, these models are not automatically downloaded. To download these models as well, use the `--all` flag while executing the script. Prior to running the script, it's necessary to request access to these models. Additionally, ensure that you are authenticated with HuggingFace by logging in through the `huggingface-cli` or by setting the `HF_TOKEN` environment variable.
```
yondonfu, line 39: I don't see anything related to an `--all` flag in `dl_checkpoints.sh`?

yondonfu (changes requested):

PR #34, opened 2024-03-13
Added LOWVRAM env variable to image-to-video

Titan-Node:
Enables image-to-video sequential cpu offloading and forward chunking, roughly 3x savings of VRAM with 5.7x longer inference time.


yondonfu (changes requested):
The same comments from https://github.com/livepeer/ai-worker/pull/36 apply here.

PR #35, opened 2024-03-13
Added LOWVRAM env variable to image-to-image

Titan-Node:
Enables image to image sequential cpu offloading, roughly 1.4x savings of VRAM with 4.6x longer inference time.


yondonfu (changes requested):
The same comments from https://github.com/livepeer/ai-worker/pull/36 apply here.

PR #40, opened 2024-03-14
feat(runner): add support for SDXL-Lightning in image-to-image pipelines

rickstaa:
This pull request adds the [SDXL-Lightning](https://huggingface.co/ByteDance/SDXL-Lightning/blob/main/sdxl_lightning_samples.jpg) model to the image-to-image pipelines.

Fixes LIV-97

### DIFF
--- a/runner/app/pipelines/image_to_image.py
+++ b/runner/app/pipelines/image_to_image.py
@@ -1,8 +1,14 @@
 from app.pipelines.base import Pipeline
 from app.pipelines.util import get_torch_device, get_model_dir
 
-from diffusers import AutoPipelineForImage2Image
-from huggingface_hub import file_download
+from diffusers import (
+    AutoPipelineForImage2Image,
+    StableDiffusionXLPipeline,
+    UNet2DConditionModel,
+    EulerDiscreteScheduler,
+)
+from safetensors.torch import load_file
+from huggingface_hub import file_download, hf_hub_download
 import torch
 import PIL
 from typing import List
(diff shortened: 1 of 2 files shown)


runner/app/pipelines/image_to_image.py
```diff
@@ -37,8 +48,49 @@ def __init__(self, model_id: str):
             kwargs["variant"] = "fp16"
 
         self.model_id = model_id
-        self.ldm = AutoPipelineForImage2Image.from_pretrained(model_id, **kwargs)
```
rickstaa, line 40: I've made some minor changes here. First, I combine the pipeline initialization and GPU allocation steps in one line. Moreover, I've transitioned to using `torch_device` variable.

runner/modal_app.py
```diff
@@ -164,6 +164,24 @@ def text_to_image_sdxl_lightning_8step_api():
     return make_api("text-to-image", "ByteDance/SDXL-Lightning-8step")
 
 
+@stub.function(image=api_image, secrets=[Secret.from_name("api-auth-token")])
```
yondonfu, line 167: The Starter plan for Modal only supports 8 deployed endpoints so as-is I think deploying this app would fail. Assuming that is the case, I would recommend removing the following endpoints to stay under the limit:

- `image_to_video_svd_api()` -> can just use the SVD1.1 endpoint

rickstaa: You're right. That's a good catch. I noticed that error yesterday when I deployed a test endpoint to my modal account. I have now commented on the SVD1.0 endpoint in 7334b3afadd80a8f97a4d1437490bfd4f80ed650 👍🏻.

yondonfu (approved):

PR #46, opened 2024-03-31
chore: add timestamp to uvicorn logs

rickstaa:
This pull request adds a uvicorn configuration file and uses this file during uvicorn startup to add timestamps to the uvicorn logs. The configuration is equal to the default configuration stored in `uvicorn.config.LOGGING_CONFIG` but with a timestamp prepended.


iameli (approved):
LGTM

PR #52, opened 2024-04-11
Add support for the `timbrooks/instruct-pix2pix` model

stronk-dev:


stronk-dev:
Since pix2pix has yet another parameter (`image_guidance_scale`), we probably want some more changes. We also need a better way to get a default `inference_steps` value. I think for most dApps they want a sensible default so they don't hjave to bother thinking about it, but it has so much influence in inference time and output quality. And the 'best' defaults differs depending on model and scheduler used

stronk-dev:
Also still needs adjustment to set "fp16" automatically

rickstaa:
Replaced by https://github.com/livepeer/ai-worker/pull/94.

rickstaa:
@stronk-dev I currently decided for transparency to use the models default. I however agree that in the future it would make sense to set more sensible defaults and enforce fp16. I created a linear item for it 👍🏻.

PR #53, opened 2024-04-11
ByteDance/AnimateDiff-Lightning text2vid

stronk-dev:
Builds on https://github.com/livepeer/ai-worker/pull/51 to add the `ByteDance/AnimateDiff-Lightning` model.
Works, but models seems very VRAM hungry. Also getting minutes of inference time vs <!0 seconds of their demo, so something needs to be finetuned.


stronk-dev:
(Wrong branch and repo)

PR #54, opened 2024-04-11
ByteDance/AnimateDiff-Lightning text2vid support

stronk-dev:
Builds on https://github.com/livepeer/ai-worker/pull/51 to add the `ByteDance/AnimateDiff-Lightning` model.
Works, but models seems very VRAM hungry. Also getting minutes of inference time vs <!0 seconds of their demo, so something needs to be finetuned.

### DIFF
--- a/runner/app/pipelines/text_to_video.py
+++ b/runner/app/pipelines/text_to_video.py
@@ -1,8 +1,9 @@
 from app.pipelines.base import Pipeline
 from app.pipelines.util import get_torch_device, get_model_dir
 
-from diffusers import DiffusionPipeline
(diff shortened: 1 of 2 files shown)


stronk-dev:
Sloppy code, but the motion LoRas are pretty neat

stronk-dev:
Note: this model seems to fall apart when generating longer videos. About 16 frames seems good. There is also a regular AnimateDiff, which should be able to hook into any (SD based) text2img model (there's a separate one for SDXL based text2img). It's much slower, but I've seen some _really_ nice results coming out of that.

PR #55, opened 2024-04-12
Experiment with combined base+refiner for SDXL

stronk-dev:
Lets not hard code it like this, maybe enhance the request to be able to request a list of sequential models?

### DIFF
--- a/runner/app/pipelines/text_to_image.py
+++ b/runner/app/pipelines/text_to_image.py
@@ -6,6 +6,7 @@ from diffusers import (
     StableDiffusionXLPipeline,
     UNet2DConditionModel,
     EulerDiscreteScheduler,
+    StableDiffusionXLImg2ImgPipeline
 )
 from safetensors.torch import load_file
 from huggingface_hub import file_download, hf_hub_download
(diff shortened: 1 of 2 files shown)


stronk-dev:
Apparently there's also a big difference between:
- Base model x Refiner:
Use the base model for the high-noise stage and the refiner for the low noise stage
```
            kwargs["denoising_end"] = 0.8
            kwargs["output_type"] = "latent"
            image = self.ldm(prompt, **kwargs).images
            del kwargs["output_type"]
            kwargs["denoising_start"] = ["denoising_end"]
            del kwargs["denoising_end"]
            kwargs["image"] = image
            return self.refiner(prompt, **kwargs).images
```

- Base model -> refiner
Use the base model to fully denoise the image and then use the refiner to add details
```
            kwargs["output_type"] = "latent"
            image = self.ldm(prompt, **kwargs).images[0]
            del kwargs["output_type"]
            kwargs["image"] = image[None, :]
            return self.refiner(prompt, **kwargs).images
```
I am getting good results with both. To support refiner models, we would require a param to set `denoising_start/end` and a separate `num_inference_steps` for the second stage. So ideally you would be able to request:
- Base model
- Base model x refiner (with `denoising_start/end`)
- Base model -> refiner

PR #56, opened 2024-04-12
Add suppport for i2vgen-xl im2vid

stronk-dev:
Added a prompt for `img2vid`, cause `i2vgen-xl` supports that!
Also added the `self.ldm.enable_vae_slicing()` tweak, otherwise the model would not run on a 20GB Vram card.
Probably good to make this a config parameter or something.


stronk-dev:
Needs adjustment to detect "fp16"

PR #61, opened 2024-04-15
feat(worker): add optimization flags

rickstaa:
This pull request enables users to pass optimization flags to the AI inference pipelines. Due to startup delays we currently only pass these flags to containers that are started warm.

### DIFF
--- a/worker/docker.go
+++ b/worker/docker.go
@@ -21,7 +21,7 @@ const containerModelDir = "/models"
 const containerPort = "8000/tcp"
 const pollingInterval = 500 * time.Millisecond
 const containerTimeout = 2 * time.Minute
-const optFlagsTimeout = 5 * time.Minute // Pipelines with optimization flags can take longer to start.
+const optFlagsContainerTimeout = 5 * time.Minute
 const containerRemoveTimeout = 30 * time.Second
 const containerCreatorLabel = "creator"
 const containerCreator = "ai-worker"
@@ -112,7 +112,7 @@ func (m *DockerManager) Borrow(ctx context.Context, pipeline, modelID string) (*
 		// The container does not exist so try to create it
 		var err error
 		// TODO: Optimization flags for dynamically loaded (borrowed) containers are not currently supported due to startup delays.
-		rc, err = m.createContainer(ctx, pipeline, modelID, false, map[string]StringBool{})
+		rc, err = m.createContainer(ctx, pipeline, modelID, false, map[string]EnvValue{})
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


worker/docker.go
```diff
@@ -70,11 +71,11 @@ func NewDockerManager(containerImageID string, gpus []string, modelDir string) (
 	}, nil
 }
 
-func (m *DockerManager) Warm(ctx context.Context, pipeline string, modelID string) error {
+func (m *DockerManager) Warm(ctx context.Context, pipeline string, modelID string, optimizationFlags OptimizationFlags) error {
```
rickstaa, line 74: @yondonfu I now decided to use the more specific `optimizationFlags` instead of the more general `flags`. We can however also change this if we think we want to bundle other container flags in the future.

worker/docker.go
```diff
@@ -110,7 +111,8 @@ func (m *DockerManager) Borrow(ctx context.Context, pipeline, modelID string) (*
 	if !ok {
 		// The container does not exist so try to create it
 		var err error
-		rc, err = m.createContainer(ctx, pipeline, modelID, false)
```
rickstaa, line 113: @yondonfu I did not yet implement the ability to use optimization flags for dynamically loaded containers since `SFAST` results in significant startup delays. We can enable them later for other optimizations like https://github.com/NVIDIA/TensorRT.

worker/docker.go
```diff
@@ -21,6 +21,7 @@ const containerModelDir = "/models"
 const containerPort = "8000/tcp"
 const pollingInterval = 500 * time.Millisecond
 const containerTimeout = 2 * time.Minute
+const optFlagsTimeout = 5 * time.Minute // Pipelines with optimization flags can take longer to start.
```
rickstaa, line 24: I just noticed that this timeout doesn't fully work. Will fix this tonight.

worker/worker.go
```diff
@@ -6,9 +6,39 @@ import (
...
+	return err
+}
+
+// String returns the string representation of the StringBool.
+func (sb StringBool) String() string {
+	return string(sb)
+}
+
+// OptimizationFlags is a map of optimization flags to be passed to the pipeline.
+type OptimizationFlags map[string]StringBool
```
yondonfu, line 40: A small nit - leave it up to you if you want to address.

IIUC the motivation for this struct is that we want to have map to store all env vars and since env var values are always strings we need to make sure if we unmarshal a JSON with a bool value that the bool value is converted into a string. If this is the case, then it seems that something like `map[string]EnvValue` might be more descriptive as a type name.

worker/docker.go
rickstaa: You're right, I overlooked that one! Thanks. I'll fix it after the watercooler chat. 👍🏻

worker/worker.go
rickstaa: Good point!

PR #63, opened 2024-04-20
Double quote variables to prevent globbing

Strykar:
This is not an issue, just good linting - https://www.shellcheck.net/wiki/SC2086


rickstaa:
@Strykar good catch! Thanks for your contribution ❤️.

Issue #49, opened 2024-04-09
Batch size too large will crash pipeline with 500 errors


ad-astra-video:
### Describe the bug

If a batch size too large for GPU is requested the container will start throwing 500 errors.  I believe the container is not crashed on this instance but is bad experience for user.

### Reproduction steps

1. Send text-to-image request with batch size of 10 to Bytedance/SDXL-Lightning model
2. ai-runner container will start throwing 500 errors as the retries continue if on RTX 4090 or lower VRAM gpu.

### Expected behaviour

O/T should be able to specify max batch size with a default of 1.  Any batch size above the set max batch size is run sequentially in the ai-runner to get the requested batch of images.  

For ByteDance/SDXL-Lightning the processing of a batch of images in one request is linear in time it takes to process the same number of images sequentially (1 takes 700ms, 3 takes 2.1s when batched together).  Its not exact but its really close so the user experience would be very similar with a batch or sequentially run image generation.  I don't expect this to be the case for all models though.  Testing and experience would drive how this works model to model but a batch size of 1 should be a safe start.

For some requests it may make sense to start returning images or making them available for download by the B as they are processed.  For fast models like ByteDance/SDXL-Lightning, probably does not need to be concerned with that.

It could also be argued, if tickets are sized to the pixels requested, the batch request should be split between multiple Os to get the batch done faster. 

### Severity

None

### OS

Linux

### Running on

Docker

### AI-worker version

latest (alpha testnet)

rickstaa:
Tracked internally at https://linear.app/livepeer-ai-spe/issue/LIV-172.

PR #66, opened 2024-04-21
Process text-to-image requested image count sequentially

ad-astra-video:
Updated the text-to-image to sequentially process images if more than 1 requested.

This provides more stable on GPU memory usage and I believe inference is relatively linear for text-to-image models.  If a user wants faster inference they can split the request to separate requests that would go to separate orchestrators.

This is a quick fix for issue [#49](https://github.com/livepeer/ai-worker/issues/49).

Some questions:
1) Should we set a sensible limit per request?  Timeout seems to be 30s so 15 sounds like a good limit that would allow most graphics cards to respond within the timeout.
2) Is there a check for if a set of images is requested but non-matching number of seeds is provided?  For example, batch of 10 requested but only 5 seeds provided.  I was not seeing where it would fill out the seeds to match the requested images count.

### DIFF
--- a/runner/app/routes/text_to_image.py
+++ b/runner/app/routes/text_to_image.py
@@ -57,24 +57,17 @@ async def text_to_image(
             ),
         )
 
-    if params.seed is None:
-        params.seed = random.randint(0, 2**32 - 1)
-        if params.num_images_per_prompt > 1:
-            params.seed = [
-                i for i in range(params.seed, params.seed + params.num_images_per_prompt)
-            ]
+    seed = params.seed if params.seed is not None else random.randint(0, 2**32 - 1)
+    seeds = [seed + i for i in range(params.num_images_per_prompt)]
 
-    if not isinstance(params.seed, list):
-        params.seed = [params.seed]
-    
-    num_images_per_prompt = params.num_images_per_prompt
-    params.num_images_per_prompt = 1
+    # TODO: Process one image at a time to avoid CUDA OEM errors. Can be removed again
+    # once LIV-243 and LIV-379 are resolved.
     images = []
     has_nsfw_concept = []
-    seeds = params.seed
-    for i in range(num_images_per_prompt):
+    params.num_images_per_prompt = 1
+    for seed in seeds:
         try:
-            params.seed = [seeds[i]] #pass one seed at a time
+            params.seed = [seed]
             imgs, nsfw_check = pipeline(**params.model_dump())
             images.extend(imgs)
             has_nsfw_concept.extend(nsfw_check)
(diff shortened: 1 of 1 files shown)


ad-astra-video:
Did some testing and when I get past 12 images, the first ones processed return an ErrNotFound from the broadcaster.  15 images returns 3 ErrNotFound, 13 returns 1 ErrNotFound.  20 returns 8 ErrNotFound.

Is there a limit of 12 somewhere? or is this just my setup?

ad-astra-video:
> Did some testing and when I get past 12 images, the first ones processed return an ErrNotFound from the broadcaster. 15 images returns 3 ErrNotFound, 13 returns 1 ErrNotFound. 20 returns 8 ErrNotFound.
> 
> Is there a limit of 12 somewhere? or is this just my setup?

I think I found it, the MemoryDriver in go-tools has a cache length of 12. Node uses a session based cache setup here https://github.com/livepeer/go-livepeer/blob/930533388e62d21b339cccd8d6f4348dc14c5e75/cmd/livepeer/starter/starter.go#L1170.  

Memory driver is in livepeer go-tools repo.  https://github.com/livepeer/go-tools/blob/ac33a3c30a694a743f0077630aaf62423b5fd1f9/drivers/local.go#L17

eliteprox:
>     1. Should we set a sensible limit per request?  Timeout seems to be 30s so 15 sounds like a good limit that would allow most graphics cards to respond within the timeout.

I think we should set a reasonable limit for the number of files sent to a given orchestrator and send remainder to the next orchestrator in the pool if one is available.

>     2. Is there a check for if a set of images is requested but non-matching number of seeds is provided?  For example, batch of 10 requested but only 5 seeds provided.  I was not seeing where it would fill out the seeds to match the requested images count.

I'll have to debug/learn the multi-file workflow first, but this line should be generating a seed when it's not provided https://github.com/livepeer/ai-worker/pull/66/files#diff-2952f66b536acb78e9bb5ee0337a00485762b802438a3209508b3b0ee088212dR62

ad-astra-video:
Updated to lastest `main` branch and force pushed to overwrite prior commits.

ad-astra-video:
> > ```
> > 1. Should we set a sensible limit per request?  Timeout seems to be 30s so 15 sounds like a good limit that would allow most graphics cards to respond within the timeout.
> > ```
> 
> I think we should set a reasonable limit for the number of files sent to a given orchestrator and send remainder to the next orchestrator in the pool if one is available.
> 
> > ```
> > 2. Is there a check for if a set of images is requested but non-matching number of seeds is provided?  For example, batch of 10 requested but only 5 seeds provided.  I was not seeing where it would fill out the seeds to match the requested images count.
> > ```
> 
> I'll have to debug/learn the multi-file workflow first, but this line should be generating a seed when it's not provided https://github.com/livepeer/ai-worker/pull/66/files#diff-2952f66b536acb78e9bb5ee0337a00485762b802438a3209508b3b0ee088212dR62

@rickstaa added a generator to create enough seeds if the images requested > 1.  

My understanding of the os memory driver is it will fail to retrieve images from the gateway if greater than 12 are requested so we have a limit at 12 I believe but no error providing that it will fail after 12 is requested.

eliteprox:
It sounds like we need to set a cap of 12 since that is the technical limit, but for reasonable processing time that might need to be even lower.

rickstaa:
@ad-astra-video merged into the main branch. Thanks a lot 🚀!

rickstaa:
> It sounds like we need to set a cap of 12 since that is the technical limit, but for reasonable processing time that might need to be even lower.

Let's deal with this in a subsequent pull request I created https://linear.app/livepeer-ai-spe/issue/LIV-380/throw-warning-when-user-requests-a-bathc-with-more-than-12-images to track this issue.

PR #69, opened 2024-04-25
feat(worker): add jpg, jpeg and .gif support in image-to-image pipeline (LIV-264)

eliteprox:
This change adds support for uploading jpg, jpeg and gif file formats to the image-to-image workflow.

### .jpeg
```
curl -X POST 0.0.0.0:8937/image-to-image -F image=@/home/elitedev/dog.jpeg -F prompt="a dog" -F model_id="ByteDance/SDXL-Lightning"
{"images":[{"seed":568555010,"url":"/stream/9aea34c2/6a3332f2.png"}]}
```
### .jpg
```
curl -X POST 0.0.0.0:8937/image-to-image -F image=@/home/elitedev/dog.jpg -F prompt="a dog" -F model_id="ByteDance/SDXL-Lightning"
{"images":[{"seed":3512476704,"url":"/stream/3b2f08fd/23daeba5.png"}]}
```
### .gif
```
curl -X POST 0.0.0.0:8937/image-to-image -F image=@/home/elitedev/sample.gif -F prompt="a sky at the beach" -F model_id="ByteDance/SDXL-Lightning"
{"images":[{"seed":4239492,"url":"/stream/507072fa/34179043.png"}]}
```

### DIFF
--- a/worker/b64.go
+++ b/worker/b64.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"fmt"
 	"image"
+	"image/gif"
+	"image/jpeg"
 	"image/png"
 	"io"
 	"os"
@@ -25,6 +27,10 @@ func ReadImageB64DataUrl(url string, w io.Writer) error {
 	switch dataURL.MediaType.ContentType() {
 	case "image/png":
 		err = png.Encode(w, img)
+	case "image/jpg", "image/jpeg":
+		err = jpeg.Encode(w, img, nil)
+	case "image/gif":
+		err = gif.Encode(w, img, nil)
 		// Add cases for other image formats if necessary
 	default:
 		return fmt.Errorf("unsupported image format: %s", dataURL.MediaType.ContentType())


rickstaa:
@eliteprox Great! This fixes https://linear.app/livepeer-ai-spe/issue/LIV-264/i2i-throws-unknown-format 🚀. 

### Why

@eliteprox given that you made changes to a function that is called at the output side of the pipeline I checked why this works while the error seems to occur before:

https://github.com/livepeer/ai-worker/blob/62ea2eef7eefb184a3f57ead27fdd81d2b8e9897/worker/b64.go#L16-L40

Is called. I think the following is going on: When no image `data_types` are defined the FastAPI api checks data types based on the [content_types](https://fastapi.tiangolo.com/reference/uploadfile/?h=content_type#fastapi.UploadFile.content_type) registered on the [Image](https://pkg.go.dev/image) package. The code you added imports:

```go
	"image/gif"
	"image/jpeg"
```

Which causes these types to be registered in the [Image](https://pkg.go.dev/image) library with allows FastAPI to deduct these image types as inputs. Additionally you now gave the pipelines the ability to output these formats as well if we want that 👍🏻.

eliteprox:
@rickstaa It was interesting that the image-to-video pipeline was already able to support jpg and gif files, but the image-to-image pipeline did not. 

Yes, this change was related to the output from the model. Even though we always return images as .png to the user, it looks like the initial encoded output from the worker's image-to-image model would matched the uploaded image type, and we only had support for png until now.

Larger files like mp4 videos are not base64 encoded, and I think that is why this only affected the image-to-image pipeline.

PR #70, opened 2024-04-25
Fix broken example code

eliteprox:
These lines were generating errors in vscode. They appear related to #61 

They are just sample code and there might be a better way to fix than using stubs. Let me know what you think.


rickstaa (approved):
LGTM! Thanks.

PR #72, opened 2024-04-28
Fix external containers

ad-astra-video:
The changes in this PR allow connection to external containers and selection of the external containers by an orchestrator more efficiently.  Limit of 1 request per container at a time initially.  A 1 request queue would be a nice addition but I think is not needed right now.

Note: the go-livepeer/ai-video for selection updates relies on this update to ai-worker.


rickstaa:
Dependent on https://github.com/livepeer/go-livepeer/pull/3033.

ad-astra-video:
Closing this. The multiple containers per pipeline/model is merged separately as well.

PR #73, opened 2024-04-29
feat(worker): add HasCapability method

rickstaa:
This commit adds a HasCapability method to the worker that can be used to check if the worker has enough capacity to take a request. This method will be used to check if an O has enough capability to support a request in https://github.com/livepeer/go-livepeer/tree/ai-video. This will prevent the Gateway from paying for requests send the the O when it has no capacity.


rickstaa:
By design this does not work with external containers (see https://github.com/livepeer/ai-worker/blob/3d355743417abde6bcae6b5b62e03a74138d4968/worker/worker.go). So people can still try to game the system they will suspended if they keep returning `insufficient capacity` errors.

PR #80, opened 2024-05-11
Add the num_inference_steps to the AI API for text-to-image endpoint

mikezupper:
This change allows higher quality image generation to the AI _text-to-image_ endpoint. The the _num_inference_steps_ parameter will be set at a sensible default of **25** (if no value is provided).

This change will require the Gateway and Orchestrator nodes to be updated.


rickstaa (approved):
@mikezupper, thanks a lot for this pull request. I reviewed it, and it works perfectly 🚀. I did, however, add one commit to handle an edge case that could crash the pipeline. Additionally, I changed the default `num_inference_steps` from `25` to `50` to match the default set by the diffusers library. Once [this issue](https://github.com/deepmap/oapi-codegen/issues/373) gets resolved, we can remove the default value from our codebase altogether.

PR #91, opened 2024-05-30
chore(runner): Pin python package versions to current version

ad-astra-video:
This PR pins the python packages to current versions as of today.  This is being done now because `setuptools` updated to 70.0.0 this week and is causing the Dockerfile build to fail.


rickstaa (approved):
LGTM thanks!

PR #93, opened 2024-05-31
feat(runner): disable DEEPCACHE for lightning/turbo models

rickstaa:
This commit ensures that people can not use the [deepcache](https://github.com/horseee/DeepCache) optimization with lightning and turbo models. As explained in https://github.com/livepeer/ai-worker/issues/82#issuecomment-2141983903 this optimization does not offer any speedup for these models while it does reduce image quality.

### DIFF
--- a/runner/app/pipelines/util.py
+++ b/runner/app/pipelines/util.py
@@ -8,6 +8,7 @@ from diffusers.pipelines.stable_diffusion import StableDiffusionSafetyChecker
 from transformers import CLIPFeatureExtractor
 from typing import Optional
 import logging
+import re
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


runner/app/pipelines/util.py
```diff
@@ -51,6 +51,28 @@ def validate_torch_device(device_name: str) -> bool:
...
+def is_lightning_model(model_id: str) -> bool:
+    """Checks if the model is a Lightning model.
+
+    Args:
+        model_id: Model ID.
+
+    Returns:
+        True if the model is a Lightning model, False otherwise.
+    """
+    return "-lightning" in model_id.lower()
```
eliteprox, line 63:
```suggestion
    return "lightning" in model_id.lower()
```
Let's use this convention for naming since some model names use both `-` and `_`

runner/app/pipelines/util.py
```diff
@@ -51,6 +51,28 @@ def validate_torch_device(device_name: str) -> bool:
...
+def is_turbo_model(model_id: str) -> bool:
+    """Checks if the model is a Turbo model.
+
+    Args:
+        model_id: Model ID.
+
+    Returns:
+        True if the model is a Turbo model, False otherwise.
+    """
+    return "-turbo" in model_id.lower()
```
eliteprox, line 74:
```suggestion
    return "turbo" in model_id.lower()
```
Let's use this convention for naming since some model names use both `-` and `_`

eliteprox (changes requested):
Tested successfully with two small changes requested

PR #116, opened 2024-07-03
add MOCK_PIPELINE env var to runner if exists

ad-astra-video:
Adds MOCK_PIPELINE to managed ai-runner container if exists.  Assists with quickly testing go-livepeer updates where pipeline results are not necessary.

### DIFF
--- a/worker/docker.go
+++ b/worker/docker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


rickstaa:
@ad-astra-video, could we use the optimization_flags to pass this environment variable to the container?

rickstaa:
> Member

Never mind I see why you would want this when people don't use managed containers 👍🏻. Let's update the patch instead.

rickstaa:
@ad-astra-video I like this addition 🚀. I updated the patch file to give people the ability to use ENV variables instead of the config file (see ea24a706d7bc42109b02ad954fd82d70166784a8). A branch in which the patch is applied is found here https://github.com/livepeer/ai-worker/pull/117. Let's close this for now since we don't want the MOCK ability in the main branch since people can abuse it.

PR #140, opened 2024-07-31
Prompt embeds update for T2I, I2I and upscale.

JJassonn69:
- [ ] the all the models worked nicely except timbrooks/pix2pix, when prompting use + or - after the word to increase the weight of the word in generation, you can add multiple ++ or -- to increase or decrease weight. 

- [ ]  the inferencing time didnot change at all most of the time or was barely reduced for some cases.

- [ ] example of prompt "a cat playing with a ball++" ,  here the ball has higher weight so it will be more focused when generating.


ad-astra-video (commented):
@jjassonn can you rebase this to current ai-worker and only include the files that changed with adding the prompt_embeds?

Compel only seems to support SD1.5 and SDXL.  Should we consider switching to sd_embed repo because it has support for SD3 and FLUX.1?  The Compel maintainer seems to have other priorities right now.

JJassonn69:
@ad-astra-video , i have cleaned up and rebased it using the main branch from ai-worker. In regards to the changing of embeddings from Compel to another is out of the scope of the bounty initially listed. We might have to add another bounty for that as it will require extensive testing with all the models and pipelines. 
initial bounty > `https://github.com/livepeer/bounties/issues/32`

runner/app/pipelines/image_to_image.py
```diff
@@ -235,12 +221,35 @@ def __call__(
                 # Default to 8step
                 kwargs["num_inference_steps"] = 8
 
+        # trying different cases for prompt_embed because some models dont support without pooled_prompt_embeds.
         try:
-            outputs = self.ldm(prompt, image=image, **kwargs)
-        except torch.cuda.OutOfMemoryError as e:
-            raise e
+            compel_proc = Compel(tokenizer=self.ldm.tokenizer, text_encoder=self.ldm.text_encoder)
```
rickstaa, line 226: @JJassonn69 I think something went wrong during a rebase. I can not find Compel in the code 🤔.

JJassonn69: compel==2.0.3 should be included in the requirements.txt, might have been removed mistakenly.

JJassonn69: fixed, included the compel in the requirements and installed the dependencies to check for any conflicts, no issues.

PR #163, opened 2024-08-17
Enable new models in audio-to-text

eliteprox:
This change adds support for new whisper models `distil-whisper/distil-large-v3` and `openai/whisper-medium`.

It also optimizes those models to use the appropriate BFLOAT, FLOAT16 or FLOAT32 values.

Credit to @ad-astra-video for intially exploring these models and optimizations


runner/app/pipelines/audio_to_text.py
```diff
@@ -40,7 +42,14 @@ def __init__(self, model_id: str):
...
-        if os.environ.get("BFLOAT16"):
+        float16_enabled = os.getenv("FLOAT16", "").strip().lower() == "true"
+        bfloat16_enabled = os.getenv("BFLOAT16", "").strip().lower() == "true"
+
+        if float16_enabled:
+            logger.info("AudioToTextPipeline using float16 precision for %s", model_id)
+            kwargs["torch_dtype"] = torch.float16
+
+        if bfloat16_enabled:
             logger.info("AudioToTextPipeline using bfloat16 precision for %s", model_id)
```
rickstaa, line 53: @eliteprox, thanks for the pull request! 🚀 It looks good overall. However, please keep in mind that the default models `openai/whisper-large-v3` and `distil-whisper/distil-large-v3` use weights in either float16 or bfloat16 formats. The `torch_dtype` parameter is primarily for the calculations during runtime. You can verify this by checking the model files in these repositories: [Hugging Face - distil-large-v3](https://huggingface.co/distil-whisper/distil-large-v3/tree/main). Notice the presence of files with the `.fp32.safetensors` extension, indicating the format being used.

If the standard `.safetensors` (fp16) format meets your needs, you might consider removing the `FLOAT16` environment variable and instead switch based on the model extension. This approach was implemented by Yondon in [this commit](https://github.com/livepeer/ai-worker/blob/557631c0a98547c1fa2c15b69afe05943e7e5838/runner/app/pipelines/audio_to_text.py#L32C9-L41). I will leave that decision to you based on your research 👍🏻. Feel free to merge when you think this pull request is done 🚀.

rickstaa (approved):

runner/app/pipelines/audio_to_text.py
eliteprox: Thanks for the tip, I updated the logic to load recommended float values by model. Tested that they download and load correctly

eliteprox:
@rickstaa I made several changes since you last reviewed this PR, so I held off on merging. Could you or @ad-astra-video re-review the latest changes?

rickstaa (approved):
@eliteprox great work! Thanks!

PR #181, opened 2024-08-29
add video support to segment-anything-2 pipeline

eliteprox:
This change updates the request field `image` to `media_file` and loads the appropriate segment-anything-2 inference model based on the content type of the file. It uses ffmpeg to process the video to image frames and loads them in with inference values from the request. Some adjustments still need to be made to the request/response parameters

### DIFF
--- a/runner/app/pipelines/segment_anything_2.py
+++ b/runner/app/pipelines/segment_anything_2.py
@@ -1,11 +1,17 @@
 import logging
+import shutil
+import tempfile
 from typing import List, Optional, Tuple
 
-import PIL
+from PIL import Image
+from fastapi import UploadFile
+import torch
 from app.pipelines.base import Pipeline
 from app.pipelines.utils import get_torch_device
 from app.routes.util import InferenceError
 from sam2.sam2_image_predictor import SAM2ImagePredictor
+from sam2.sam2_video_predictor import SAM2VideoPredictor
+import subprocess
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


runner/app/pipelines/segment_anything_2.py
```diff
@@ -15,22 +21,73 @@ def __init__(self, model_id: str):
...
-            raise InferenceError(original_exception=e)
+                _, out_obj_ids, out_mask_logits = self.tm_vid.add_new_points_or_box(
+                    inference_state,
+                    frame_idx=5,
+                    obj_id=1,
+                    points=kwargs.get('point_coords', None),
+                    labels=kwargs.get('point_labels', None),
+                    )
+                video_segments = {}  # video_segments contains the per-frame segmentation results
+                for out_frame_idx, out_obj_ids, out_mask_logits in self.tm_vid.propagate_in_video(inference_state):
```
pschroedl, line 83: I have a feeling we should return the full triple instead of creating a video segment, leaving post-processing to the consumer of the API ( though I recognize this is good quick way to validate the sanity of the mask outputs )

eliteprox: Thanks for the guidance on this, I see how we can just return the results of `self.tm_vid.propagate_in_video(inference_state):`

eliteprox: > I have a feeling we should return the full triple instead of creating a video segment, leaving post-processing to the consumer of the API ( though I recognize this is good quick way to validate the sanity of the mask outputs )

I had some issues trying to return the correct values. I added frame index as an input parameter, normally `propagate_in_video` will loop returning results for each frame starting at the frame index until the end of the video, now it should only return a single frame. But the data doesn't look correct, can you take a look?

PR #188, opened 2024-09-04
refactor(runner): add InferenceError to all pipelines

rickstaa:
This pull request adds the inference error logic from the SAM2 pipeline to all pipelines so users are given a warning when they supply wrong arguments. Please note that the 422 thrown by FastAPI are not handled corretly and we don't do validation at the go-livepeer side yet (see https://linear.app/livepeer-ai/project/improve-error-handling-117c11ff78f3/issues). As a result although this pull request improves the behavoir we don't yet have an optimised solution.


runner/app/routes/audio_to_text.py
```diff
@@ -14,39 +15,47 @@
...
         e
     ) or isinstance(e, AudioConversionError):
         status_code = status.HTTP_415_UNSUPPORTED_MEDIA_TYPE
         error_message = "Unsupported audio format or malformed file."
+    elif "CUDA out of memory" in str(e) or isinstance(e, OutOfMemoryError):
+        status_code = status.HTTP_400_BAD_REQUEST
+        error_message = "Out of memory error."
+    elif isinstance(e, InferenceError):
+        status_code = status.HTTP_400_BAD_REQUEST
+        error_message = str(e)
```
rickstaa, line 39: Check if errors are forwarded to the client.

rickstaa: Just commited my stashed changes to this pull request. Should be fixed now. The only error that is not correctly handled yet is the 422 error. We should add validation at the go-livepeer side (See https://linear.app/livepeer-ai/issue/AI-628/correctly-handle-422-errors) and add a FastAPI validation error exception handler.

runner/app/routes/image_to_video.py
```diff
@@ -17,6 +19,35 @@
...
+def handle_pipeline_error(e: Exception) -> JSONResponse:
+    """Handles exceptions raised during image-to-video pipeline processing.
+
+    Args:
+        e: The exception raised during image-to-video processing.
+
+    Returns:
+        A JSONResponse with the appropriate error message and status code.
+    """
+    if isinstance(e, torch.cuda.OutOfMemoryError):
```
leszko, line 32: I see this part of the code is repeated in every pipeline, would it maybe make sense to extract it and reuse?

leszko (approved):
Added one comment, other than that LGTM

runner/app/routes/image_to_video.py
rickstaa: @leszko as also stated in the discord I implemented a new variant of @gioelecerati's logic in ae8df749e107cf2e13ab8fb86f5a884675370e1c.

PR #190, opened 2024-09-05
trim worker err handling code

rickstaa:
This pull request restructures the worker error handling logic so that the code is a bit leaner.


worker/worker.go
```diff
@@ -219,34 +206,25 @@ func (w *Worker) Upscale(ctx context.Context, req UpscaleMultipartRequestBody) (
 		return nil, err
 	}
 
-	if resp.JSON422 != nil {
-		val, err := json.Marshal(resp.JSON422)
```
rickstaa, line 223: Check why this was used before.

PR #192, opened 2024-09-11
chore(ci): add 'base' tag and segment-anything-2 docker image build

ad-astra-video:
Updates github docker action to add `livepeer/ai-runner:base` image tag and builds the segment-anything-2 image.

The segment-anything-2 dockerfile was also updated to move the base image to a ARG line that can be overriden using `--build-arg` to set the base image.


rickstaa:
@ad-astra-video thanks! Replaced with https://github.com/livepeer/ai-worker/pull/193 so that the action runs.

PR #216, opened 2024-10-01
Added example_data folder to Dockerfile

Titan-Node:
Dockerfile is missing `example_data` folder for bench.py to run. Built and tested the fix with `titannode/ai-runner:example-images`, works as expected.


eliteprox (approved):
Makes it easier to benchmark pipelines that require file input without needing to use a volume mount. LGTM.

PR #217, opened 2024-10-01
Pin sam2 commit

pschroedl:
Pins the dependency on [facebook/sam2](https://github.com/facebookresearch/sam2) to unbreak our sam2 pipeline.

With our workflow pulling from upstream main, we encountered:
```
    raise MissingConfigException(
hydra.errors.MissingConfigException: Cannot find primary config 'configs/sam2/sam2_hiera_l.yaml'. Check that it's in your config search path.
```

Others have reported this issue as well:
https://github.com/facebookresearch/sam2/issues/346


eliteprox (approved):
LGTM, thanks for fixing the build! This is a temporary fix until the sam-2 repository is updated with a solution

PR #226, opened 2024-10-11
worker: return back error strings to O

gioelecerati:
Modifying worker.go:

- Appending the error message to the string that is currently getting sent back to the O and the `Gateway`, so that they can log the proper reasons of failure on the pipelines.

Modifying runner:

- Added an helper under `util.py` in the routes to handle pipeline exception. In this helper, known Exceptions or known patterns can return a humanized `str` as error and the specified status code for the exception. Otherwise, they return either a default error message and status code or the specified error string
- Potentially we could remove this helper and instead have this logic on the Gateway, to have humanized strings to be returned via api response to the requester.
- Currently not edited the llm pipeline, as I see it's returning dicts instead of string. I was unable to check if it's returning dict for a reason or not, to be updated later if we want to handle specific exceptions

### DIFF
--- a/runner/app/routes/image_to_image.py
+++ b/runner/app/routes/image_to_image.py
@@ -7,7 +7,7 @@ import torch
 from app.dependencies import get_pipeline
 from app.pipelines.base import Pipeline
 from app.pipelines.utils.utils import LoraLoadingError
-from app.routes.util import HTTPError, ImageResponse, http_error, image_to_data_url
+from app.routes.util import HTTPError, ImageResponse, http_error, image_to_data_url, handle_pipeline_exception
 from fastapi import APIRouter, Depends, File, Form, UploadFile, status
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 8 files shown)


rickstaa:
@gioelecerati thanks for your pull request. Unfortunatelly, this was already implemented in [PR #188](https://github.com/livepeer/ai-worker/pull/188). I had some last-minute changes stashed on my local machine, which I’ve just committed, and I also cleaned up the code a bit. I really like how you combined the `handle_pipeline_exception` into a single function—maybe you could apply that logic on top of my changes?

Just keep in mind that I still have several other changes related to error handling that need to be reviewed (see [PR #190](https://github.com/livepeer/ai-worker/pull/190)). The current error handling is far from optimal, and it’s part of a broader project I’ve been working on to improve it (details [here](https://linear.app/livepeer-ai/project/improve-error-handling-117c11ff78f3/issues)).

rickstaa:
@gioelecerati I included your logic in https://github.com/livepeer/ai-worker/pull/188/commits and made you the author of that commit.

PR #227, opened 2024-10-15
fix torch device comparison

yaodingyd:
[torch.device](https://pytorch.org/docs/stable/tensor_attributes.html#torch-device) is a class not a string, so directly compare them will also result in True

```
torch_device != "cpu" # this is always True, event when torch_device is torch.device('cpu')
```


rickstaa (approved):
Very good catch. Thanks!

PR #228, opened 2024-10-15
(feat) add return_timestamps as configurable in request

eliteprox:
This change adds the `return_timestamps` parameter to the audio-to-text pipeline, allowing end-users to configure the inference job to return timestamps at word-level, sentence-level or no timestamps at all.

Supported values for `return_timestamps` are `false` and `word`. The pipeline defaults to the current sentence-level timestamp transcription to avoid breaking changes with existing applications.


runner/app/pipelines/audio_to_text.py
```diff
@@ -20,7 +20,7 @@
 
 
 class AudioToTextPipeline(Pipeline):
-    def __init__(self, model_id: str):
+    def __init__(self, model_id: str, return_timestamps: str = "false"):
```
rickstaa, line 23: @eliteprox variable appears to be unused.

runner/app/pipelines/audio_to_text.py
```diff
@@ -77,11 +76,19 @@ def __call__(self, audio: UploadFile, **kwargs) -> List[File]:
             converted_bytes = audio_converter.convert(audio, "mp3")
             audio_converter.write_bytes_to_file(converted_bytes, audio)
 
+        if kwargs["return_timestamps"] != "false":
```
rickstaa, line 79: @eliteprox can you investigate in the [transformers library](https://github.com/huggingface/transformers) what options they have. Do they just have true and false? If so we can also just use those options.

runner/app/pipelines/audio_to_text.py
```diff
@@ -77,11 +76,19 @@ def __call__(self, audio: UploadFile, **kwargs) -> List[File]:
...
+        else:
+            kwargs.pop("return_timestamps", None)
+
         try:
             outputs = self.tm(audio.file.read(), **kwargs)
         except Exception as e:
             raise InferenceError(original_exception=e)
 
+        # When timestamps are not requested, add the chunks field to the response to satisfy the response schema
+        if "chunks" not in outputs:
```
rickstaa, line 90: I think you could use `s.pop("return_timestamps", None)`.

eliteprox: This is related to the output. When `return_timestamps` is not used, the `chunks` property is not in the output (ex: `{'text': ""}`). This logic appends the chunks property to satisfy the response schema in TextResponse.

runner/app/pipelines/audio_to_text.py
eliteprox: Removed unused variable

runner/app/pipelines/audio_to_text.py
rickstaa: @eliteprox I took a look at the behavior, and I believe it's best to keep it consistent with the current model, even though the interface has double typing. I've made these adjustments and merged the pull request. Thanks for the great addition! 🚀

eliteprox: Nice improvement with `Union[str, bool]`, I didn't know there was a way to accept two different types in one request parameter

PR #233, opened 2024-10-22
Add live_video_to_video pipeline

emranemran:
Base work for the Realtime Video AI Pipeline


j0sh:
Is the pipeline 'video-to-video' ? I have been calling it 'live-video-to-video' - see https://github.com/livepeer/go-livepeer/pull/3210

Can the main branch be main? Easier to track and keep up to date, and it seems there is a lot of stuff in the `feature/live-ai` that we do not need

j0sh:
Fixed docker paths in https://github.com/livepeer/ai-worker/pull/233/commits/b302de95022a14aa586a26c3586e283465185490 @emranemran - the commit showed up as you since it was from your machine

leszko (approved):
Good enough to merge into `main` and work from there. Also it does not touch any existing code, so nothing should break :)

j0sh:
I like the flattening of `app/python/infer` into `app/infer`. Everything in this repo involves inference though... how about `app/live` since that describes the use case?

leszko:
> I like the flattening of `app/python/infer` into `app/infer`. Everything in this repo involves inference though... how about `app/live` since that describes the use case?

I think it's a good idea. Let me rename it

PR #245, opened 2024-10-29
runner: force output buffers to flush logs immediately using PYTHONUN…

emranemran:
Logs are missing from subprocesses probably because the buffers are only flushed once full and may be lost if the subprocess crashes -- so force output buffers to flush logs immediately using `PYTHONUNBUFFERED` env var.


leszko (approved):

leszko:
LGTM, but let's wait for @j0sh to comment if if solves his problem.

j0sh:
This does not work. We need this: https://github.com/livepeer/ai-worker/pull/254

leszko:
Closing this one, since @j0sh sent another PR

PR #252, opened 2024-10-30
Add publish / subscribe params for live video-to-video

j0sh:
A bit of groundwork for the rest


runner/app/pipelines/live_video_to_video.py
```diff
@@ -34,12 +34,12 @@ def __call__(
                     input_address="tcp://localhost:5555",
                     output_address="tcp://localhost:5556",
                     http_port=8888,
+                    subscribe_url=kwargs["subscribe_url"],
+                    publish_url=kwargs["publish_url"],
                     initial_params=json.dumps(kwargs),
```
victorges, line 39: Since we will have other stuff in the `kwargs`, can you change this to be `kwargs["params"]` instead? otherwise we will be sending publish/subscribe URLs as params to the models, which will explode if we're validating `additionalFields` there (I think [we might](https://github.com/livepeer/ai-worker/blob/3e2b1ee7cf422dd7d269d877e332cb8637a4d9d2/runner/app/live/pipelines/streamkohaku.py#L13))

j0sh: Fixed in https://github.com/livepeer/ai-worker/pull/252/commits/1b9efc2e0b197b44d81b29f8db30fdbb19ca96a9 but this may blow up in the ZeroMQStreamer if `params` is not a dict, eg 

```
curl -vvv http://localhost:9000/live-video-to-video/ -H 'Content-Type: application/json' -d '{"publish_url":"abc","subscribe_url":"def", "params":"aaaaa"}'
```

should be 

```
curl -vvv http://localhost:9000/live-video-to-video/ -H 'Content-Type: application/json' -d '{"publish_url":"abc","subscribe_url":"def", "params":{"aaaaa":"bbbb"}}'
```

(omitting the params is fine too)

https://github.com/livepeer/ai-worker/blob/ac556e8163b1855feb162799eaf70e87eaef5fc4/runner/app/live/infer.py#L20

victorges: Params has to be a dict!

victorges: we can enforce that on the schema maybe or add an explicit check for `object` here perhaps

j0sh: objectified in https://github.com/livepeer/ai-worker/pull/252/commits/80a036a099ca704b262f8dc0611acf5ad72bb33e

victorges (approved):

PR #253, opened 2024-10-30
live/infer: Add trickle input and output

j0sh:
Logging is still rough as I'm ironing out issues around tear-down, but wanted to put this out

Should be testable with 

```
 curl -vvv http://localhost:9000/live-video-to-video/ -H 'Content-Type: application/json' -d '{"subscribe_url":"http://trickle-server/input-stream","publish_url":"http://trickle-server/output-stream"}'
```

### DIFF
--- a/runner/app/live/infer.py
+++ b/runner/app/live/infer.py
@@ -86,7 +86,10 @@ if __name__ == "__main__":
         logging.error(f"Error parsing --initial-params: {e}")
         sys.exit(1)
 
-    logging.basicConfig(level=logging.INFO)
+    logging.basicConfig(
+        format='%(asctime)s %(levelname)-8s %(message)s',
+        level=logging.INFO,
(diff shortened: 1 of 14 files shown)


j0sh:
@victorges Just a heads up that the publish task hangs after the input stream completes. I am not sure exactly how the other parts work but you will need to call `publish_queue.put(None)` after receiving a None from `subscribe_queue` (or waiting for subscribe task completion, eg `await subscribe_task` that is created [here](https://github.com/livepeer/ai-worker/pull/253/files#diff-7cc27ec3a97f82590d84c1bca16961c3e2681d486716eeaf3698760d106d446bR28). This can also be used as a signal to clean up the inference job.

For reference, [here](https://gist.github.com/j0sh/7ec40d2d1c7108c62203adaf07107430#file-main-py) is how I have been driving the process (without inference)

PR #265, opened 2024-11-06
runner/live: Cleanup dockerfiles

victorges:
Several simplifications made.
- Removed duplciate python install
- Removed NVIDIA packages that already come in base image (or actually should be host only)
- Removed intermediate multimedia image and kept only base
- Moved python installation to the streamdiffusion image (each pipeline has their own)
- Made the live-apps image a little more configurable, it will be reused on top of different pipelines
- Probably more stuff

Actual image size didn't reduce much, from 27.4GB to 27.1GB (yeah streamdiffusion is huge and maybe it's downloading models), but I think it will be much easier to work with these now. Also removed all the redundant installations I could find.


j0sh (approved):
🚢 🇮🇹

PR #267, opened 2024-11-08
live/infer: Allow running with zeromq protocol

victorges:
I'll update the `livepeer/captions-server` LiveKit PoC to use this repository as a submodule.
So making the version from `infer.py` here compatible with running the zeromq streaming
protocol.


varshith15 (approved):
LGTM

PR #269, opened 2024-11-08
Add ComfyUI pipeline

leszko:
WIP

### DIFF
--- a/runner/app/live/pipelines/comfyui.py
+++ b/runner/app/live/pipelines/comfyui.py
@@ -1,0 +1,19 @@
+from comfystream.client import ComfyStreamClient
+
+class ComfyPipeline(Pipeline):
+  def __init__(self, **params):
+    # configure as is needed
+    self.client = ComfyStreamClient(...)
+    self.update_params(**params)
+
+  async def process_frame(self, image: Image.Image) -> Image.Image:
+    # preprocess image into a torch.Tensor
+    output = self.client.queue_prompt(input)
+    # postprocess output into a Image.Image
+    return output
+
+  def update_params(self, **params):
+    # Convert params into a Prompt type which describes the workflow
+    self.client.set_prompt(prompt)
+
+  ...
--- a/runner/docker/Dockerfile.live-base-comfyui
+++ b/runner/docker/Dockerfile.live-base-comfyui
@@ -1,0 +1,54 @@
+ARG BASE_IMAGE=livepeer/ai-runner:live-base
+FROM ${BASE_IMAGE}
+
+RUN apt update && apt install -y vim
+
+# Create directory for ComfyUI custom nodes and models
+RUN mkdir -p /comfyui/custom_nodes
+
+# Install required Python version
+ARG PYTHON_VERSION=3.10
+RUN pyenv install $PYTHON_VERSION && \
+    pyenv global $PYTHON_VERSION && \
+    pyenv rehash
+
+# Upgrade pip and install required packages
+ARG PIP_VERSION=23.3.2
+ENV PIP_PREFER_BINARY=1
+RUN pip install --no-cache-dir --upgrade pip==${PIP_VERSION} setuptools==69.5.1 wheel==0.43.0
+
+# Install ComfyUI-Depth-Anything-Tensorrt Node (https://github.com/yuvraj108c/ComfyUI-Depth-Anything-Tensorrt)
+RUN cd /comfyui/custom_nodes && \
+    git clone https://github.com/yuvraj108c/ComfyUI-Depth-Anything-Tensorrt.git && \
+    cd ComfyUI-Depth-Anything-Tensorrt && \
+    pip install -r requirements.txt
+
+# Upgrade TensorRT to 10.6.0
+RUN pip uninstall -y tensorrt && \ 
+    pip install tensorrt==10.6.0
+
+WORKDIR /app
+
+RUN pip install torch torchvision torchaudio tqdm
+
+# Temp install specific versions of opentelemetry package before they merge this PR: https://github.com/hiddenswitch/ComfyUI/pull/26
+RUN pip install opentelemetry-distro==0.48b0 && \
+    pip install opentelemetry-exporter-otlp==1.27.0 && \
+    pip install opentelemetry-propagator-jaeger==1.27.0 && \
+    pip install opentelemetry-instrumentation==0.48b0 && \
+    pip install opentelemetry-util-http==0.48b0 && \
+    pip install opentelemetry-instrumentation-aio-pika==0.48b0 && \
+    pip install opentelemetry-instrumentation-requests==0.48b0 && \
+    pip install opentelemetry-semantic-conventions==0.48b0
+
+# Install comfystream (which includes ComfyUI)
+RUN pip install git+https://github.com/yondonfu/comfystream.git 
+
+RUN git clone https://github.com/yondonfu/comfystream.git && \
+    cd comfystream && \
+    pip install -r requirements.txt && \
+    cd ..
+
+WORKDIR /app/comfystream
+
+CMD ["python", "server/app.py", "--workspace", "/comfyui", "--port", "3389", "--host", "0.0.0.0"]
--- a/runner/docker/README.md
+++ b/runner/docker/README.md
@@ -35,18 +35,28 @@ To build a pipeline-specific container, you need to build the base container fir
 
    This command builds the `segment-anything-2` pipeline-specific container using the Dockerfile located at [docker/Dockerfile.segment_anything_2](docker/Dockerfile.segment_anything_2) and tags it as `livepeer/ai-runner:segment-anything-2`.
 
-### Steps to Build a Realtime Video AI Container
+   ### ComfyStream integration with Depth-Anything
 
-   ```bash
-   docker build -t livepeer/ai-runner:live-base . -f docker/Dockerfile.live-base
-   docker build -t livepeer/ai-runner:live-multimedia -f docker/Dockerfile.live-multimedia .
-   docker build -t livepeer/ai-runner:live-stream-diffusion -f docker/Dockerfile.live-stream-diffusion .
-   docker build -t livepeer/ai-runner:live-apps -f docker/Dockerfile.live-apps .
+   1. Build Docker image
+   ```
+   docker build -t livepeer/ai-runner:live-base-comfyui -f docker/Dockerfile.live-base-comfyui .
    ```
 
-   Then, you can run and test the Live Container with the following commands:
-   ```bash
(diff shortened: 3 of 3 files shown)


runner/app/live/pipelines/comfyui.py
```diff
@@ -0,0 +1,63 @@
...
+
+  def process_frame(self, image: Image.Image) -> Image.Image:
+    # transform = transforms.ToTensor()
+    transform = transforms.Compose([
+      transforms.Resize((512, 512)),  # Resize the image to 512x512
+      transforms.ToTensor(),          # Convert the image to a tensor
+      transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])  # Normalize the image
+    ])
+
+    # Apply the transformation to convert to torch.Tensor
```
yondonfu, line 41: There is a simpler way to convert a PIL.Image into a BHWC (ComfyUI expects this [shape](https://docs.comfy.org/essentials/custom_node_tensors#what-is-a-tensor)) torch.Tensor.

```
# Convert PIL.Image to numpy
# Normalize by dividing by 255 to ensure the tensor values are between 0 and 1
image_np = np.array(image).astype(np.float32) / 255.0
# Convert from numpy to torch.Tensor
# Initially, the torch.Tensor will have shape HWC but we want BHWC
# unsqueeze(0) will add a batch dimension at the beginning of 1 which means we just have 1 image
image_tensor = torch.tensor(image_np).unsqueeze(0)
```

A few other examples of this type of conversion for reference:
- The example server converts [from numpy to torch.Tensor](https://github.com/yondonfu/comfystream/blob/d6791a8d38d3b6953a340021eeda610ebc7dc530/server/pipeline.py#L19)
- The diffusers codebase contains helper functions like [pil_to_numpy](https://github.com/huggingface/diffusers/blob/5b972fbd6a6c50cf1afdf1ba34c34d84fc67861c/src/diffusers/image_processor.py#L157) and [numpy_to_pt][(https://github.com/huggingface/diffusers/blob/5b972fbd6a6c50cf1afdf1ba34c34d84fc67861c/src/diffusers/image_processor.py#L177).

runner/app/live/pipelines/comfyui.py
```diff
@@ -0,0 +1,63 @@
...
+    if image.mode != 'RGB':
+      image = image.convert('RGB')
+    tensor = transform(image)
+    if tensor.dim() == 3:  # If there's no batch dimension
+      tensor = tensor.unsqueeze(0)
+    conv_layer = nn.Conv2d(3, 512, kernel_size=1)
+    extended_tensor = conv_layer(tensor)
```
yondonfu, line 48:
```
RuntimeError: The size of tensor a (3) must match the size of tensor b (512) at non-singleton dimension 1
```

FWIW this error is thrown by this line because the shape of `tensor` is incompatible with layer created by `nn.Conv2d()`. `conv_layer()` will expect a BCHW shape tensor, but `tensor` is a BHWC shape tensor. Technically, you could fix this by reshaping `tensor` via something like:

```
# BCHW -> BHWC
tensor = tensor.permute(0, 3, 1, 2)
```

But, as I mentioned in my other comment, I think you can probably skip this!

runner/app/live/pipelines/comfyui.py
leszko: Nice, it worked. I also had to convert the image to "RGB", but then it started to work. Thanks Yondon!

runner/app/live/pipelines/comfyui.py
leszko: Yeah, I think it's not needed. I guess I was just trying out some random things to get it working 🙃 Thanks for the help!

runner/requirements.live-ai.txt
```diff
@@ -7,7 +6,6 @@ Pillow==10.3.0
 python-multipart==0.0.9
 uvicorn==0.30.0
 huggingface_hub==0.23.2
-xformers==0.0.23
```
leszko, line 10: @victorges @j0sh I had to remove a few dependencies here and added them into `Dockerfile.base-streamdiffusion`. Please check if it does not break liveportrait and streamdiffusion pipeline dependencies.

runner/app/live/pipelines/comfyui.py
```diff
@@ -0,0 +1,78 @@
...
+    comfy_ui_workspace = os.getenv(COMFY_UI_WORKSPACE_ENV)
+    self.client = ComfyStreamClient(cwd=comfy_ui_workspace)
+
+    params = {'prompt': json.loads(DEFAULT_WORKFLOW_JSON)}
+    self.update_params(**params)
+
+    # Comfy will cache nodes that only need to be run once (i.e. a node that loads model weights)
+    # We can run the prompt once before actual inputs come in to "warmup"
+    warmup_input = torch.randn(1, 512, 512, 3)
+    asyncio.get_event_loop().run_until_complete(self.client.queue_prompt(warmup_input))
```
victorges, line 57: - Is there a way not to use `asyncio` here? We run these pipelines on separate processes for minimum overhead, and I think ideally we should avoid these libs that alter the way the code runs as it could affect the models performance (had great performance gains by removing the pipeline logic from the main process that uses asyncio etc)
- WDYT of making a more explicit management of the event loop here, since we're not running on a process that is guaranteed to have one set up? Meaning: 
  - creating an event loop on construction with `asyncio.new_event_loop`
  - keeping a reference to it in the class and reusing it on `process_frame`
  - closing the event loop on the `__del__` function when the object gets garbage collected (apparently python does ref counting and this will happen as soon as we recreate the pipeline for example). could also do any other necessary cleanup there like closing the comfyui client

victorges, line 57: Do we need to warmup only on startup or anytime the params change? I'd consider moving this to either `update_params` or in `process_frame` with a similar `self.first_frame` var, like the streamdiffusion pipeline (soft preference for the latter to avoid heavy processing on a constructor).

runner/docker/README.md
```diff
@@ -35,18 +35,30 @@ To build a pipeline-specific container, you need to build the base container fir
...
-
-   curl --location -H "Content-Type: application/json" 'http://localhost:8000/live-video-to-video' -X POST -d '{"stream_url":"http://<url-to-trickle-pull>"}'
-   ```
\ No newline at end of file
+### ComfyStream integration with Depth-Anything
+
+1. Build Docker image
+```
+export PIPELINE=comfyui
+docker build -t livepeer/ai-runner:live-base-comfyui -f docker/Dockerfile.live-base-comfyui .
```
victorges, line 43:
```suggestion
docker build -t livepeer/ai-runner:live-base-${PIPELINE} -f docker/Dockerfile.live-base-${PIPELINE} .
```

runner/requirements.live-ai.txt
victorges: Makes sense to me. I'm out of a dev machine for now but will test later.

victorges (approved):
LGTM!

runner/requirements.live-ai.txt
eliteprox: We also needed to remove xformers from here and install it within our base file for segment-anything-2

runner/app/live/pipelines/comfyui.py
yondonfu: For context, [ComfyStreamClient.queue_prompt()](https://github.com/yondonfu/comfystream/blob/34bea08e3cd16fc1e0e4cc566961df881450b1f2/src/comfystream/client.py#L21) is currently async out of convenience - it [awaits a call to the async function of ComfyUI library](https://github.com/hiddenswitch/ComfyUI/blob/44be2591df18d31e2e751e589b70f62d0317cc0b/comfy/client/embedded_comfy_client.py#L191) and a future that references the result of prompt execution. If the async version can be used for now, I think that would be the simplest path forward vs. creating a non-async variant.

@victorges IIUC if queue_prompt() is run via run_until_complete() then that just ends up mostly equivalent to calling a non-async version of queue_prompt() and while there may be some extra overhead from event loop management that should be pretty small relative to the actual computation happening in queue_prompt(). Do you have specific reason to believe that there would actually be larger perf degradation? I know you saw perf gains by moving inference code into a separate process without async code, but it is not clear to me whether the general usage of async code or specific, inefficient async code was the cause of perf degradation in that case.

Either way if the Depth-Anything-V2 workflow can run fine in real-time for now then IMO can always figure out how to further optimize separately!

yondonfu: The "warmup" probably only needs to happen on startup to ensure that the one-time expensive operations, like loading model weights, in the workflow are run before the first real input arrives. If the param values for any of the nodes in the workflow changes, then those nodes would be re-executed the next time the workflow is run.

Re: avoiding heavy processing in a constructor - for reference, in the pipeline for the comfystream example server, I perform this "warmup" operation in a [separate function](https://github.com/yondonfu/comfystream/blob/34bea08e3cd16fc1e0e4cc566961df881450b1f2/server/pipeline.py#L14) that ultimately just executes the workflow under the hood.

runner/requirements.live-ai.txt
leszko: Thanks 👍

runner/app/live/pipelines/comfyui.py
leszko: Yeah, I think that unless we see clear benefits of `asyncio.new_event_loop`, then I'd keep it as it is for now. We can optimize later. Anyway, Victor, interested in your thoughts.

leszko: Yeah, I think that we should either extend the "interface" to also include `warm` there (as used by Yondon). Otherwise, I think it's ok to keep it in the constructor.

yondonfu (approved):
Feel free to react to/respond to [this comment](https://github.com/livepeer/ai-worker/pull/269#discussion_r1838535353) as you see fit.

runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -0,0 +1,50 @@
...
+    pip install -r requirements.txt
+
+# Upgrade TensorRT to 10.6.0
+RUN pip uninstall -y tensorrt && \ 
+    pip install tensorrt==10.6.0
+
+RUN pip install torch==2.5.1 torchvision torchaudio tqdm
+
+# Temp install specific versions of opentelemetry package before they merge this PR: https://github.com/hiddenswitch/ComfyUI/pull/26
+RUN pip install opentelemetry-distro==0.48b0 && \
```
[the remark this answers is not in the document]
leszko: Updated in d2fa0829882d19de763a84550301bab085fc6d7e

runner/requirements.live-ai.txt
victorges: FTR needed a small fix for this: https://github.com/livepeer/ai-worker/pull/277

PR #271, opened 2024-11-09
live/trickle: Updates for go-livepeer

j0sh:
Use `Lp-Trickle-Seq` as the header for sequence numbering

Better end-of-stream handling

Ignore SSL certificates since orchestrators generate them at runtime

Don't assume a fixed frame rate for output video; use what we have

Logging cleanup

### DIFF
--- a/runner/app/live/trickle/trickle_subscriber.py
+++ b/runner/app/live/trickle/trickle_subscriber.py
@@ -47,11 +47,11 @@ class TrickleSubscriber:
                 self.pending_get = await self.preconnect()
 
             # Extract the current connection to use for reading
-            conn = self.pending_get
+            resp = self.pending_get
             self.pending_get = None
 
             # Extract and set the next index from the response headers
-            segment = Segment(conn)
+            segment = Segment(resp)
 
             if segment.eos():
                 return None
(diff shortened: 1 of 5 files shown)


runner/app/live/trickle/trickle_subscriber.py
```diff
@@ -62,14 +51,19 @@ async def next(self):
             self.pending_get = None
 
             # Extract and set the next index from the response headers
-            idx = await self.get_index(conn)
-            if idx != -1:
+            segment = Segment(conn)
```
victorges, line 54: Is this `conn` var actually a response obj? Got slightly confused that it's passed as `conn` and becomes `resp(onse)` on these functions (old `get_index` or `Segment` constructor). Might be worth renaming here

victorges (approved):

runner/app/live/trickle/trickle_subscriber.py
j0sh: They are the same, renamed to `resp` for consistency in https://github.com/livepeer/ai-worker/pull/271/commits/8b411dc94000f71d3e4bbcafa9224a8b782dd00a

PR #272, opened 2024-11-11
runner/live: Make sure to resize images outside of infer process

victorges:
On the LiveKit PoC we resized the images on the agent code, and were missing that part from this infer.py piece. This is probably one reason for the inference to be lower on the ai-worker as the resize still happened, but it happened on the same process as we run the model so it was competing for the same Python GIL (and resizing on CPU might be some relevant time).

### DIFF
--- a/runner/app/live/streamer/streamer.py
+++ b/runner/app/live/streamer/streamer.py
@@ -121,7 +121,15 @@ class PipelineStreamer(ABC):
             async for frame in self.ingress_loop(done):
                 if done.is_set() or not self.process:
                     return
-                logging.debug(f"Sending input frame. width: {frame.width}, height: {frame.height}")
+
+                # crop the max square from the center of the image and scale to 512x512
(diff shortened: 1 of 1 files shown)


j0sh:
We already resize as part of the [video-to-image](https://github.com/livepeer/ai-worker/blob/main/runner/app/live/trickle/media.py#L83) flow. Do we need to re-process it here?

victorges:
@j0sh hmm I see. bad news that this doesn't explain the worse performance then :3

I think it might still be worth having it to have this logic "self-contained" and not rely on an external logic to work (it doesn't just get slow, it actually explodes when using tensorrt).

Since I also have the checks to not reprocess in case it has the right size, I think it's beneficial to have it. WDYT?

victorges:
@j0sh yeah I don't think we should preserve the aspect ratio. The input has to be 512x512

PR #273, opened 2024-11-12
feat: add hardware info reporting

ad-astra-video:
This PR adds routes for `/hardware/info` and `/hardware/stats` to enable reporting hardware information to go-livepeer ai-worker.  Hardware information is polled by the ai-worker at the startup using the `/hardware/info` route and added to the `RunnerContainer`.  It can also be used by Orchestrators to monitor their GPUs periodically.

To provide the GPU information in logs, this PR prints the cuda devices information at startup of the runner.

The `/hardware/stats` route is setup to use with additional build out in go-livepeer to check runtime stats as models are processing.  Note that right now, the `pipeline` call blocks the runner so the stats route cannot execute while the pipeline call is in process.  The pipeline call can be moved to non blocking by adding `await asyncio.to_thread(pipeline, **kwargs)` to the pipeline call.


rickstaa:
@ad-astra-vide0 please checkout https://github.com/ad-astra-video/ai-worker/pull/1 and let me know what you think. I think with those changes we can merge 👍🏻.

rickstaa (approved):
LGTM after https://github.com/ad-astra-video/ai-worker/pull/1 merge.

rickstaa (approved):

PR #275, opened 2024-11-14
worker+runner: Dynamically start live inference runners

victorges:
This implements support on the worker to dynamically start the runner containers
and manage their lifecycle.

This is slightly different from how the existing APIs work, since the runner container
runs long past the original request that started it. The solution was:
- Just start the container on the request, but don't return it automatically
- Implement a logic on the runner to stop the process when the input stops
- Implement a monitor on the worker to track when the container stops
- Update the worker's docker manager state when the container stop to allow other containers to run

I haven't been able to test this yet because my dev machine exploded. I'll create
a new one to try this out.


j0sh (approved):
🚢 🚢 🚢

runner/app/pipelines/base.py
```diff
@@ -5,8 +5,9 @@
 class Pipeline(ABC):
     @abstractmethod
     def __init__(self, model_id: str, model_dir: str):
+        self.model_id: str # type hint so we can use the field in routes
```
leszko, line 8: what does this comment mean?

```
# type hint so we can use the field in routes
```

leszko (approved):
LGTM

worker/docker.go
```diff
@@ -275,6 +285,8 @@ func (m *DockerManager) createContainer(ctx context.Context, pipeline string, mo
 	m.containers[containerName] = rc
 	m.gpuContainers[gpu] = containerName
 
+	go m.watchContainer(ctx, rc)
```
rickstaa, line 288: @victorges I like your changes! However, I think there’s an advantage to keeping containers warm for a certain amount of time for batch jobs if requests continue to come in. This could help improve response times and resource efficiency.

victorges: I see! So we should not remove the container when the context is cancelled right? I can change the logic to just return the container on that case instead WDYT? With that I can even remove the "return container" function that is currently called explicitly, and instead just let the callers pass a ctx that is cancelled when the request is done and the container is auto-returned.

runner/app/pipelines/base.py
victorges: This avoids a typing error on all the `route/` implementations when they used the `model_id` field on the abstract `Pipeline` class (currently tons of typing errors on this proj, I'll try to clear them up as I make changes). 

I'm changing this comment to this which might be clearer?
```
# declare the field here so the type hint is available when using this abstract class
```

worker/docker.go
victorges: Implemented on https://github.com/livepeer/ai-worker/pull/275/commits/4d5adbef48f81fc60489daf24367d8e452319dde !

mjh1 (approved):

PR #276, opened 2024-11-14
Add Noop pipeline for local testing

leszko:


rickstaa:
Like this I think this can replace the mock patch -> https://github.com/livepeer/ai-worker/blob/main/runner/dev/README.md#mocking-the-pipelines?

j0sh (approved):
LGTM, do you think its also worth applying some light transformation to the image (eg, inverting it) so it is doing _something_? Might be a thing to control too, eg to toggle back and forth.

But then again, it wouldn't be a "noop" anymore if we did that :)

leszko:
> LGTM, do you think its also worth applying some light transformation to the image (eg, inverting it) so it is doing _something_? Might be a thing to control too, eg to toggle back and forth.
> 
> But then again, it wouldn't be a "noop" anymore if we did that :)

Hehe, yeah, I'd keep it just noop for the sake of simplicity. Will just add some logging params are updated.

victorges (commented):
Neat!

PR #278, opened 2024-11-15
runner: Create liveportrait pipeline container

victorges:
The main goal for this PR was to create the liveportrait pipeline docker image.

On the way to that, ended up doing a couple improvements also to the `streamdiffusion` pipeline:
- Now we mount the directory where `streamdiffusion` lib builds the TensorRT models, so we don't need to do it on every boot
- Created a mode for the `dl_checkpoints.sh --tensorrt` script to already pre-build the live pipelines as TensoRT models


runner/dl_checkpoints.sh
```diff
@@ -79,7 +79,40 @@ function download_all_models() {
...
+}
+
+function build_tensorrt_models() {
+    download_all_models
+
+    printf "\nBuilding TensorRT models...\n"
+
+    BUILD_TRT='python app/live/StreamDiffusionWrapper/build_tensorrt.py'
+    docker run --rm -it -v ./models:/models --gpus all \
+        livepeer/ai-runner:live-app-streamdiffusion \
```
eliteprox, line 93:
```suggestion
        livepeer/ai-runner:live-app-streamdiffusion \
        -e PIPELINE=live-video-to-video -e MODEL_ID=streamdiffusion \
```
I had to include these two variables for the container to stay running while tensorrt builds. We probably need similar variables for LivePortrait too

runner/dl_checkpoints.sh
```diff
@@ -79,7 +79,42 @@ function download_all_models() {
...
+function build_tensorrt_models() {
+    download_all_models
+
+    printf "\nBuilding TensorRT models...\n"
+
+    # Matrix of models and timesteps to compile StreamDiffusion TensorRT engines for.
+    MODELS="stabilityai/sd-turbo KBlueLeaf/kohaku-v2.1"
+    TIMESTEPS="3 4" # This is basically the supported sizes for the t_index_list
+    docker run --rm -it -v ./models:/models --gpus all \
+        livepeer/ai-runner:live-app-streamdiffusion \
```
eliteprox, line 95:
```suggestion
        livepeer/ai-runner:live-app-streamdiffusion \
        -e PIPELINE=live-video-to-video -e MODEL_ID=streamdiffusion \
```
I had to include these two env variables for the container to stay running while tensorrt is compiled

victorges: I actually don't think those envs would be necessary since the scripts that we run in here don't check those variables (they are very simple `tensorrt` compilation scripts). Did you notice that I changed the command that I run within the container when starting them here? Were you doing that as well when you ran these?

If you were running exactly this script and still got the errors, can you share the terminal output here?

eliteprox: I see now, sorry I was copying this into a separate script so to avoid the `download_all_models` and I lost those variables. Instead of downloading all models, maybe we can just download the ones we're building tensorrt for?

runner/dl_checkpoints.sh
```diff
@@ -79,7 +79,42 @@ function download_all_models() {
 
     # Download live-video-to-video models.
     huggingface-cli download KBlueLeaf/kohaku-v2.1 --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
-    huggingface-cli download KwaiVGI/LivePortrait --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
+    huggingface-cli download stabilityai/sd-turbo --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
+    huggingface-cli download warmshao/FasterLivePortrait --local-dir models/FasterLivePortrait--checkpoints
+}
+
+function build_tensorrt_models() {
```
leszko, line 86: I wonder if we want to have one function to download and build all models. Is that what Orchestrators do, do they download ALL models?

Maybe we should have some granularity, like to download only the models we use for live video?

runner/dl_checkpoints.sh
```diff
@@ -79,7 +79,42 @@ function download_all_models() {
...
+        livepeer/ai-runner:live-app-streamdiffusion \
+        bash -c "for model in $MODELS; do
+                    for timestep in $TIMESTEPS; do
+                        echo \"Building TensorRT engines for model=\$model timestep=\$timestep...\" && \
+                        python app/live/StreamDiffusionWrapper/build_tensorrt.py --model-id \$model --timesteps \$timestep
+                    done
+                done"
+
+    docker run --rm -it -v ./models:/models --gpus all \
+        livepeer/ai-runner:live-app-liveportrait \
```
leszko, line 104: Will you need to first build docker images? Or will they be hosted in Docker Hub?

runner/dl_checkpoints.sh
```diff
@@ -79,7 +79,42 @@ function download_all_models() {
 
     # Download live-video-to-video models.
     huggingface-cli download KBlueLeaf/kohaku-v2.1 --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
-    huggingface-cli download KwaiVGI/LivePortrait --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
+    huggingface-cli download stabilityai/sd-turbo --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
+    huggingface-cli download warmshao/FasterLivePortrait --local-dir models/FasterLivePortrait--checkpoints
```
leszko, line 83:
```suggestion
    huggingface-cli download warmshao/FasterLivePortrait --local-dir models/FasterLivePortrait--checkpoints
    huggingface-cli download yuvraj108c/Depth-Anything-Onnx --include depth_anything_vitl14.onnx --local-dir models
```
Would you mind adding also Depth Anything here? I had it in README, but this script is a better place I guess.

runner/dl_checkpoints.sh
```diff
@@ -79,7 +79,42 @@ function download_all_models() {
...
+                        sh scripts/all_onnx2trt.sh
+                    else
+                        echo 'Regular LivePortrait TensorRT engines already exist, skipping build'
+                    fi && \
+                    if [ ! -f '/models/FasterLivePortrait--checkpoints/liveportrait_animal_onnx/stitching_lip.trt' ]; then
+                        echo 'Building TensorRT engines for LivePortrait models (animal)...'
+                        sh scripts/all_onnx2trt_animal.sh
+                    else
+                        echo 'Animal LivePortrait TensorRT engines already exist, skipping build'
+                    fi"
```
leszko, line 117:
```suggestion
                    fi"

    # DepthAnything for ComfyUI
    docker run --rm -it -v ./models:/models --gpus all \
        livepeer/ai-runner:live-app-comfyui \
        bash -c "cd /models; python /comfyui/custom_nodes/ComfyUI-Depth-Anything-Tensorrt/export_trt.py"
    mkdir -p ./models/tensorrt/depth-anything
    mv ./models/*.engine ./models/tensorrt/depth-anything
```

leszko (approved):
LGTM

runner/dl_checkpoints.sh
victorges: Yeah I've been thinking we could have some kind of "tags" here, so you can download models only related to a specific pipeline or sth. I'll keep this as a separate improvement tho for the future, wdyt

runner/dl_checkpoints.sh
victorges: They will be on docker hub! For now you have to build them though

runner/dl_checkpoints.sh
victorges: I tried taking a look on the readme instructions but I think they will need some rethinking on the way the comfyui nodes are setup. I don't think we should use `--local-dir models` for example as that will download files to the root of the cache directory which would get messy. This likely needs its own sub directory as well similar to the faster live portrait and StreamDiffusion folders, but then the compilation script and docker image need to be updated too. If that ok I'll leave that refactor to a follow up PR for you, which has context on the comfyui image. Happy to review that

leszko: Ok, let's leave it as a separate PR.

runner/dl_checkpoints.sh
leszko: Yeah, let's keep it as a separate PR

victorges:
@leszko implemented the ComfyUI stuff here if you could take a look: https://github.com/livepeer/ai-worker/pull/284

PR #286, opened 2024-11-22
(lv2v) change control_url to not nullable

eliteprox:
This change modifies the `control_url` field to not-nullable. This change goes along with https://github.com/livepeer/go-livepeer/pull/3263


runner/app/routes/live_video_to_video.py
```diff
@@ -118,6 +120,13 @@ async def live_video_to_video(
             ),
         )
 
+    if "trickle_port" in params.params:
+        trickle_port = params.params["trickle_port"]
```
leszko, line 124: Why do you need this? Shouldn't we pass correct port as part of `subscribe_url`/`publish_url`/`control_url`? I think that if we need to change it we should change it in O.

leszko (changes requested):

runner/app/routes/live_video_to_video.py
eliteprox: We can pass the port at the beginning of the URL instead and then append the host within the runner conditionally by checking for the colon and port number. I was trying to be more descriptive by explicitly passing the port with a named parameter, but I think either approach is good

leszko: To me it's super weird that you pass Url and then separately the port. So I'd prefer to not do that logic unless really needed.

leszko: Or maybe I'm missing something. But Afaiu all urls can already include the port so no need to pass it separately

eliteprox: The URL is passed without a host or port (example: `/ai/trickle/2d4a3F`). By using `request.client.host` we are able to reliably identify the host to use for following trickle updates, regardless of the serviceUri config of the orchestrator. This is why the block is added. However, I think we can use `:8936/ai/trickle/2d4a3F`, and append host when the value is passed starting with `:` if you'd like it a bit cleaner. This still allows for testing with a full URL.

eliteprox: Modified this to remove the `trickle_port` param entirely. It looks like we're going with ServiceURI. In the case a local development, an external host IP will need to be used.

PR #291, opened 2024-11-27
Allow running worker on non-GPU machine

leszko:


rickstaa:
@leszko looks good. How would this be used. One reason we didn't include these things in the codebase is that it makes it easy for people to put a mocked orch on the network.

rickstaa:
We have used patches to do these kind of things in the past https://github.com/livepeer/ai-worker/tree/main/runner/dev/patches.

worker/docker.go
```diff
@@ -472,3 +474,14 @@ tickerLoop:
...
+
+func portOffset(gpu string) string {
+	if isEmulatedGPU(gpu) {
+		return strings.Replace(gpu, "emulated-", "", 1)
+	}
+	return gpu
+}
+
+func isEmulatedGPU(gpu string) bool {
+	return strings.HasPrefix(gpu, "emulated-")
```
victorges, line 486: How do you set one of these emulated ones?

victorges (approved):

PR #293, opened 2024-11-28
Support for JSON string in defaultImage field for overriding pipeline specific images in the mappings 

RUFFY-369:
Towards closing ['livepeer/bounties/#69](https://github.com/livepeer/bounties/issues/69)

This PR adds support for processing JSON string in `defaultImage`and specifies pipeline-specific images as key-value pairs with their mappings being updated to override `containerImage`.

`go-livepeer` PR which it supports is [#3284](https://github.com/livepeer/go-livepeer/pull/3284)

cc @rickstaa

### DIFF
--- a/worker/docker.go
+++ b/worker/docker.go
@@ -103,6 +103,23 @@ type DockerManager struct {
 	mu         *sync.Mutex
 }
 
+// updatePipelineMappings updates the specified mapping with pipeline to image overriding.
+// It logs a warning if a pipeline is not found in the given mapping.
+//
+// Parameters:
+// - overrides: A map of pipeline names to custom image names.
+// - mapping: The map to be updated with the provided overrides.
+// - mapName: The name of the map (used for logging purposes).
+func updatePipelineMappings(overrides map[string]string, mapping map[string]string, mapName string) {
+	for pipeline, image := range overrides {
+		if _, exists := mapping[pipeline]; exists {
+			mapping[pipeline] = image
+		} else {
+			slog.Warn("Pipeline not found in map", "map", mapName, "pipeline", pipeline)
+		}
+	}
+}
+
 // overridePipelineImages function parses a JSON string containing pipeline-to-image mappings and overrides the default mappings if valid. 
 // It updates the `pipelineToImage` and `livePipelineToImage` maps with custom images.
 // Parameters:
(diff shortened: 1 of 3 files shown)


worker/docker.go
```diff
@@ -97,6 +97,33 @@ type DockerManager struct {
 	mu         *sync.Mutex
 }
 
+func overridePipelineImages(defaultImage string) error {
```
rickstaa, line 100: @RUFFY-369 could you add a basic function docsting and tests?

RUFFY-369: @rickstaa Thanks for the review! I just added the docstring and the test in the latest commits. :+1: :rocket:

worker/docker.go
```diff
@@ -57,20 +57,57 @@ var containerHostPorts = map[string]string{
...
 
+// overridePipelineImages updates base and pipeline images with the provided overrides.
+func overridePipelineImages(imageOverrides string) error {
+	if imageOverrides == "" {
+		return fmt.Errorf("empty string is not a valid image override")
+	}
+
+	// Handle JSON format for multiple pipeline images.
+	var imageMap map[string]string
+	if err := json.Unmarshal([]byte(imageOverrides), &imageMap); err == nil {
```
victorges, line 82: WDYT of implementing a simple "json flag" abstrraction like [this](https://github.com/livepeer/stream-tester/pull/327/files#diff-2c056ea791c910800601e0619e71cdf8f40a60b4108227c309ba5252eff77ea4R52)? There are some advantages:
- Clearer error handling. If the JSON is bad it will fail on startup, not here
- Cleaner logic here, no need to handle parsing
- Minor: A bit more performant since we're not parsing the flag every time

victorges: Edit: Just saw that this inline edits the consts declared in this file. I'd say that's a sub-ideal since we now have consts that are not really the values that might be used in runtime. Some of the advantages I made above don't make sense as I guess this logic is run only once, so nevermind the JSON flag if you stay with this approach.

OTOH I think mutating the static variables can be a little error prone and harder to understand/debug. We now have "consts" in the file that are not really what might be used in runtime. A cleaner logic IMO would be to hold on to the override map in the docker manager instead, and change the [`getContainerImageName`](https://github.com/livepeer/ai-worker/blob/cb9f788a9a92409212d37b12fc09c716949030e6/worker/docker.go#L215) function to handle the override map instead. I think the logic might be simpler to understand as well. WDYT?

worker/docker.go
```diff
@@ -57,20 +57,57 @@ var containerHostPorts = map[string]string{
...
+			if pipeline == "base" {
+				defaultBaseImage = image
+				continue
+			}
+
+			// Check and update the pipeline images.
+			if _, exists := pipelineToImage[pipeline]; exists {
+				pipelineToImage[pipeline] = image
+			} else if _, exists := livePipelineToImage[pipeline]; exists {
+				livePipelineToImage[pipeline] = image
```
victorges, line 93: I think the override JSON format could be improved to avoid multiple conflicts here. We have 3 different identifiers as keys on the same map: 
- the special `base` image name which is used as the default here
- the batch pipeline image names (which comes from the `pipeline_id`)
- the live pipeline image names (which comes from the `model_id`)

All of these values could conflict with each other. I think we could have a layered JSON and avoid this, like:
```
type ImageOverrides struct {
    Default string
    Batch map[string]string
    Live map[string]string
}
```
Then we access each of them independently. WDYT?

victorges (approved):
LGTM as it works, but I suggested some code changes. I'll open a separate branch to share with you to facilitate.

victorges:
@rickstaa here's a branch with my suggested changes: https://github.com/livepeer/ai-worker/compare/aca9cc8...vg/update/ai-runner-image-flag-support

Keep in mind I only updated `docker.go`, but not the tests. If you agree with the change would you mind updating the tests as well?

worker/docker.go
rickstaa: Love it 😍. I did briefly have just this but then removed it again to keep changes minimal.

rickstaa:
> @rickstaa here's a branch with my suggested changes: [aca9cc8...vg/update/ai-runner-image-flag-support](https://github.com/livepeer/ai-worker/compare/aca9cc8...vg/update/ai-runner-image-flag-support)
> 
> Keep in mind I only updated `docker.go`, but not the tests. If you agree with the change would you mind updating the tests as well?

Thanks will do!

worker/docker.go
rickstaa: Good point and solution thanks!

victorges (approved):

PR #295, opened 2024-11-29
ci(runner): add runner test ci

rickstaa:
This pull request adds a new GitHub action that runs the runner tests on main branch commits and pull requests.

### DIFF
--- a/.github/workflows/ai-runner-test.yaml
+++ b/.github/workflows/ai-runner-test.yaml
@@ -1,0 +1,34 @@
+name: Run AI Runner Tests
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


leszko (commented):
What's the difference between this PR and https://github.com/livepeer/ai-worker/pull/296 ? 🤔

rickstaa:
@leszko this pull request adds runner tests which are python based. The other adds worker tests which are golang based. We could combine them into one CI but I kept them separate since we might want to move the worker code into `go-livepeer`.

leszko (approved):

PR #301, opened 2024-12-02
fix: managed containers return after inference completed

ad-astra-video:
@victorges @rickstaa Please review this PR to fix container return for managed containers inference.  

This was found from a report from an Orchestrator getting `insufficient capacity` error when using managed containers after the first request was completed.

I confirmed this was happening on my test machine and also confirmed it was happening on `ImageToImage` pipeline.

This fix delays the `return` in the `Borrow` function until after the `go m.watchContainer(rc, ctx)` is started.

### DIFF
--- a/worker/docker.go
+++ b/worker/docker.go
@@ -174,7 +174,6 @@ func (m *DockerManager) Borrow(ctx context.Context, pipeline, modelID string) (*
 
 	for _, runner := range m.containers {
 		if runner.Pipeline == pipeline && runner.ModelID == modelID {
-			delete(m.containers, runner.Name)
 			rc = runner
 			break
 		}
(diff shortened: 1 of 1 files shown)


worker/docker.go
```diff
@@ -169,24 +169,29 @@ func (m *DockerManager) Stop(ctx context.Context) error {
 func (m *DockerManager) Borrow(ctx context.Context, pipeline, modelID string) (*RunnerContainer, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	var rc *RunnerContainer
+	var err error
 
 	for _, runner := range m.containers {
 		if runner.Pipeline == pipeline && runner.ModelID == modelID {
 			delete(m.containers, runner.Name)
```
victorges, line 177: I guess we can remove this line now since it's done on L189
```suggestion
```

victorges (approved):
nice catch 😮‍💨

PR #306, opened 2024-12-04
runner/live: Refactor pipeline streamer and add some docs

victorges:
This does a few slightly unrelated refactors to improve general code quality and maintainability on
the live runner code. Specifically:
- Make the "protocols" be used by composition instead of inheritance.
- Remove the old `t_index_ratio_list` which was only helpful back when I was doing updates manually on a JSON
- Add some documentation to the `Pipeline` base class, helpful for pipeline implementors


eliteprox (approved):
LGTM! It's a nice cleanup and improves logging and documentation. Tested start/update stream from gateway

PR #308, opened 2024-12-04
feat: add flash_attention_2 to audio_to_text pipeline's attn_implementation

RUFFY-369:
This PR enables flash attention for `audio-to-text` pipeline. 
`runner/docker/Dockerfile.audio_to_text` can be used to build image for flash_attention_2 enabled A2T pipeline. `sdpa` attention implementation has been added in fallback condition for the GPUs that don't support flash attention. `sdpa` also improves the latency.

### DIFF
--- a/runner/app/pipelines/audio_to_text.py
+++ b/runner/app/pipelines/audio_to_text.py
@@ -5,7 +5,6 @@ from enum import Enum
 from typing import List
 
 import torch
-import warnings
 from fastapi import File, UploadFile
 from transformers import AutoModelForSpeechSeq2Seq, AutoProcessor, pipeline
 
@@ -74,7 +73,7 @@ class AudioToTextPipeline(Pipeline):
         if (major, minor) >= (8, 0):
             attn_implementation = "flash_attention_2"
         else:
-            warnings.warn(f"GPU {device.name} (Compute Capability {major}.{minor}) is not compatible with FlashAttention so, scaled_dot_product_attention is being used instead")
+            logger.warning(f"GPU {device.name} (Compute Capability {major}.{minor}) is not compatible with FlashAttention, so scaled_dot_product_attention is being used instead.")
             attn_implementation = "sdpa"
 
         # Get model specific configuration parameters.
--- a/runner/app/pipelines/audio_to_text.py
+++ b/runner/app/pipelines/audio_to_text.py
@@ -5,6 +5,7 @@ from enum import Enum
 from typing import List
 
 import torch
+import warnings
 from fastapi import File, UploadFile
 from transformers import AutoModelForSpeechSeq2Seq, AutoProcessor, pipeline
 
@@ -65,6 +66,17 @@ class AudioToTextPipeline(Pipeline):
 
         torch_device = get_torch_device()
 
+        # Get the GPU device properties
+        device = torch.cuda.get_device_properties(0)
+        major, minor = device.major, device.minor
+
+        # Compute capability of Ampere starts at 8.0
+        if (major, minor) >= (8, 0):
+            attn_implementation = "flash_attention_2"
+        else:
+            warnings.warn(f"GPU {device.name} (Compute Capability {major}.{minor}) is not compatible with FlashAttention so, scaled_dot_product_attention is being used instead")
+            attn_implementation = "sdpa"
+
         # Get model specific configuration parameters.
         model_enum = ModelName.from_value(model_id)
         self._model_cfg: ModelConfig = MODEL_CONFIGS.get(model_enum, ModelConfig())
@@ -81,7 +93,8 @@ class AudioToTextPipeline(Pipeline):
             low_cpu_mem_usage=True,
             use_safetensors=True,
             cache_dir=get_model_dir(),
-            attn_implementation="eager",  # TODO: enable flash attention.
+            torch_dtype="auto",
+            attn_implementation=attn_implementation,
             **kwargs,
         ).to(torch_device)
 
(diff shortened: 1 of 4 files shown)


runner/app/pipelines/audio_to_text.py
```diff
@@ -81,7 +93,7 @@ def __init__(self, model_id: str):
             low_cpu_mem_usage=True,
             use_safetensors=True,
             cache_dir=get_model_dir(),
-            attn_implementation="eager",  # TODO: enable flash attention.
+            attn_implementation=attn_implementation,
             **kwargs,
         ).to(torch_device)
```
eliteprox, line 99: There is a runtime warning on startup:
> You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with `model.to('cuda')`.

Loading the model to the CPU first, and then GPU appeared to clear it
```suggestion
            low_cpu_mem_usage=True,
            use_safetensors=True,
            cache_dir=get_model_dir(),
            **kwargs,
        ).to('cpu')

        # Move the model to the GPU
        model = model.to(torch_device)
        model.config.attn_implementation = attn_implementation
```

eliteprox (commented):
Great work, thanks for the PR! Here is one suggested improvement

runner/app/pipelines/audio_to_text.py
RUFFY-369: @eliteprox Thanks for the suggested change. Commited the suggestion :rocket:

runner/app/pipelines/audio_to_text.py
```diff
@@ -65,6 +66,17 @@ def __init__(self, model_id: str):
...
 
+        # Get the GPU device properties
+        device = torch.cuda.get_device_properties(0)
+        major, minor = device.major, device.minor
+
+        # Compute capability of Ampere starts at 8.0
+        if (major, minor) >= (8, 0):
+            attn_implementation = "flash_attention_2"
+        else:
+            warnings.warn(f"GPU {device.name} (Compute Capability {major}.{minor}) is not compatible with FlashAttention so, scaled_dot_product_attention is being used instead")
```
rickstaa, line 77: @RUFFY-369 is there any reason to use the warnings package over logger.warn?

RUFFY-369: @rickstaa Not a specific reason but just followed basic code habit that I have for using warnings package for issuing warning which devs can find useful as logger.warn will log it for the user but it isn't that critical for a log so left it out just as a warning display with warnings package.

If you want I can add this warning as a log :+1:

rickstaa: I'll check it tomorrow. It's a good habit, and I suggested it mainly to maintain consistency in the code.

RUFFY-369: Sure, then let me just log in the warning as it would be more in line with the code pattern of all the other pipeline files too. :+1:

runner/app/pipelines/audio_to_text.py
rickstaa: Strange this issue should not exist if cuda_device is GPU.

RUFFY-369: I think its something with Flash attention in Transformers model as there were some issues answered where this warning disappeared by using the method in the above suggested changes.

runner/app/pipelines/audio_to_text.py
```diff
@@ -65,16 +65,23 @@ def __init__(self, model_id: str):
...
-        # Get the GPU device properties
-        device = torch.cuda.get_device_properties(0)
-        major, minor = device.major, device.minor
-
-        # Compute capability of Ampere starts at 8.0
-        if (major, minor) >= (8, 0):
-            attn_implementation = "flash_attention_2"
+        # Enable FlashAttention based on device compatibility.
+        attn_implementation = "eager"
+        if torch_device.type == "cuda":
```
rickstaa, line 70: @RUFFY-369 I added some code to ensure that CPU inference still works.

runner/app/pipelines/audio_to_text.py
```diff
@@ -92,12 +99,10 @@ def __init__(self, model_id: str):
             low_cpu_mem_usage=True,
             use_safetensors=True,
             cache_dir=get_model_dir(),
+            attn_implementation=attn_implementation,
+            device_map={"": torch_device},
```
rickstaa, line 103: @RUFFY-369, @eliteprox This updated code ensures correct device placement while eliminating the need to call `.to` multiple times.

runner/app/pipelines/audio_to_text.py
```diff
@@ -107,7 +112,7 @@ def __init__(self, model_id: str):
             tokenizer=processor.tokenizer,
             feature_extractor=processor.feature_extractor,
             max_new_tokens=128,
-            device=torch_device,
+            device_map={"", torch_device},
```
rickstaa, line 115: @RUFFY-369, @eliteprox I had to change this from device to device_maps to ensure all parts of the inference workflow ere on the right device.

rickstaa (approved):
@RUFFY-369 Looks good to me! Please review my latest changes, and if everything checks out, go ahead and merge. 🙏🏻

runner/app/pipelines/audio_to_text.py
RUFFY-369: Oh! Thanks for adding it. :+1:

runner/app/pipelines/audio_to_text.py
RUFFY-369: :eyes: :+1:

runner/app/pipelines/audio_to_text.py
rickstaa: Applied https://github.com/livepeer/ai-worker/pull/308/commits/f05dfce45ca4d750ca6162f12636188713d41477.

PR #309, opened 2024-12-04
runner/live: Report pipeline status on trickle channel

victorges:
This is to implement the logic to report the status of the live pipeline as it is running.
It uses a trickle channel to publish status updates periodically including all the metrics
that we currently have from the pipeline. 

On the process of implementing, also refactored a couple of stuff, like moving the control
subscribe loop to inside the trickle protocol as well.

The high-level logic is:
- The process.py has been instrumented to collect logs and errors from the sub-process and expose them to the streamer with some helper functions
- Start a new background task from the streamer to report status (notice that this one runs continuously. It is not killed when the process is killed, like the other tasks)
- This loop will iterate every 10s and collect the metrics from a status field and send them over trickle
- It is also triggered immediately when there are errors on the process, it is restarted, or params updated
- Notice that errors are picked up from the monitor loop periodically, so we don't risk overflowing status updates on crash loops


runner/app/live/streamer/process.py
```diff
@@ -24,10 +24,10 @@ def __init__(self, pipeline_name: str):
         self.ctx = mp.get_context("spawn")
 
         self.input_queue = self.ctx.Queue(maxsize=5)
-        self.last_input_time = 0.0
         self.output_queue = self.ctx.Queue()
-        self.last_output_time = 0.0
         self.param_update_queue = self.ctx.Queue()
+        self.error_queue = self.ctx.Queue()
+        self.log_queue = self.ctx.Queue(maxsize=100)  # Keep last 100 log lines
```
gioelecerati, line 30: Just so I understand, will this populate log_queue with an array of the last 100 log lines? @ecmulli just want your opinion on this, is this something we need in the event and with so much data?

gioelecerati (approved):

runner/app/live/streamer/process.py
victorges: We keep 100 logs here but actually only send the last 10 lines in the event. I think at some point we should be streaming all the logs tho, that's what fal.ai does too. It will be useful for any lower-level developer access that we give to write pipelines (there are some comfy nodes to print stuff for example, useful for debugging).

Also, as I mentioned, restarts are way more common than errors, and on the restart case the logs are the only thing we can provide. So they are very useful IMO and a must-keep.

PR #310, opened 2024-12-05
Fix missing nvidia-ml-py

leszko:


runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -25,7 +25,7 @@ RUN cd /comfyui/custom_nodes && \
 RUN pip uninstall -y tensorrt && \
     pip install tensorrt==10.6.0
 
-RUN pip install torch==2.5.1 torchvision torchaudio tqdm
+RUN pip install torch==2.5.1 torchvision torchaudio tqdm nvidia-ml-py==12.560.30 \
```
pwilczynskiclearcode, line 28: few lines below we install `-r requirements.txt`.
Why do we install manually and using `-r` at the same time?

pwilczynskiclearcode (approved):

mjh1 (approved):

runner/docker/Dockerfile.live-base-comfyui
eliteprox: This was already solved in #305 using requirements files. Both options are valid, but we should go with one other the other. https://github.com/livepeer/ai-worker/commit/dbefe4201a7153983479a6d15bdfaf46687c8330

leszko: I believe it didn't work with my fix. You can check, but the images were failing for us in staging.

rickstaa: @leszko could you check if https://github.com/livepeer/ai-worker/pull/319/files works? If not then there is a dependency chain preventing the right version to be installed and we shuold only specify it in the docker container.

PR #313, opened 2024-12-05
runner/live: Ignore keepalive messages from control subscriber

j0sh:
The control channel only sees sporadic messages, so the publisher will send down a keepalive periodically to prevent clients from timing out and to prevent the server from sweeping idle channels.

Needed for https://github.com/livepeer/go-livepeer/pull/3296 and https://github.com/livepeer/go-livepeer/pull/3299


victorges (approved):
neat

PR #314, opened 2024-12-06
(chore) fix startup with asyncio.create_task

eliteprox:
This change fixes an issue with the comfyui pipelines starting inference. It uses `asyncio.create_task(streamer.wait()))` instead of `streamer.wait()`.

Tested with managed/external containers and with both python 3.10/11


victorges (approved):

rickstaa:
Good find 🚀!

PR #315, opened 2024-12-06
fix: slow resize

varshith15:
The difference is almost 10x (PIL vs cv2)

Screenshot_2024-12-06_at_1 06 29_PM

Images are very similar:

Screenshot 2024-12-06 at 10 30 28 PM


runner/app/live/streamer/streamer.py
```diff
@@ -142,12 +144,19 @@ async def run_ingress_loop(self, done: Event):
 
                 # crop the max square from the center of the image and scale to 512x512
                 # most models expect this size especially when using tensorrt
-                width, height = frame.size
+                frame_array = np.array(frame)
```
victorges, line 147: Can you avoid doing any processing if resize is not necessary? so maybe keep cropping on pillow, which is cheap, and move the opencv conversion stuff to inside that resize branch. wdyt?

In production setup, the trickle client may already resize the image for us. At some point this should be a pipeline-specific behavior so this logic should be here anyway (and makes simpler to write other clients like the e2e test you're building). So we should avoid doing unnecessary pillow->cv->pillow if no resize is required here.

runner/app/live/streamer/streamer.py
```diff
@@ -142,12 +144,19 @@ async def run_ingress_loop(self, done: Event):
...
+                height, width = frame_array.shape[:2]
+
                 if width != height:
                     square_size = min(width, height)
-                    frame = frame.crop((width // 2 - square_size // 2, height // 2 - square_size // 2, width // 2 + square_size // 2, height // 2 + square_size // 2))
-                if frame.size != (512, 512):
-                    frame = frame.resize((512, 512))
+                    start_x = width // 2 - square_size // 2
+                    start_y = height // 2 - square_size // 2
```
victorges, line 153: oh i like these haha

victorges (approved):
LGTM either way as I think the extraneous processing should be relatively cheap

runner/app/live/streamer/streamer.py
varshith15: takes less than 1ms that conversion, but good catch!

victorges (approved):
neat!

PR #316, opened 2024-12-06
Live video to video/comfyui sam2

pschroedl:
Adds AI SPE realtime custom_nodes:

ComfyUI-SAM2-Realtime (https://github.com/pschroedl/ComfyUI-SAM2-Realtime.git)

ComfyUI-StreamDiffusion (https://github.com/pschroedl/ComfyUI-StreamDiffusion.git)

as well as additional nodes for effects:

ComfyUI_RyanOnTheInside (https://github.com/ryanontheinside/ComfyUI_RyanOnTheInside.git)

ComfyUI-Misc-Effects ( Starburst ) (https://github.com/ryanontheinside/ComfyUI-Misc-Effects.git)


seanhanca (approved):
approved. 

is ryanontheinside needed for the SAM2 and SD? or used for something else? can you elaborate a bit more?

runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -21,6 +21,27 @@ RUN cd /comfyui/custom_nodes && \
...
+RUN git clone https://github.com/pschroedl/ComfyUI-StreamDiffusion.git && \
+    cd ComfyUI-StreamDiffusion && \
+    pip install -r requirements.txt && \
+    cd ..
+
+# Install ComfyUI-SAM2-Realtime (https://github.com/pschroedl/ComfyUI-SAM2-Realtime.git)
+RUN git clone -b relax_python_3_10 https://github.com/pschroedl/ComfyUI-SAM2-Realtime.git && \
+    cd ComfyUI-SAM2-Realtime && \
+    pip install -r requirements.txt && \
+    cd ..
```
victorges, line 34: FTR the `RUN` commands from docker instantiate their own shell. Changing directory will not change the directory for the next docker commands. So no need to restore the previous workdir!

pschroedl: Good heads up - removed those and made sure we were cloning each into the correct directory ( could use WORKDIR I suppose

pschroedl:
> approved.
> 
> is ryanontheinside needed for the SAM2 and SD? or used for something else? can you elaborate a bit more?

They will be used in documentation for compositing using SAM2 masks and other computationally inexpensive transformations.

pschroedl:
Image build succeeded but the check failed saving to cache
`ERROR: failed to solve: error writing manifest blob: failed commit on ref "sha256:5cb82258bc2995ab335dd25f408478367e70b0ee0454c3b5f22d5a64331ae79b": unexpected status from PUT request to https://registry-1.docker.io/v2/***/build/manifests/cache: 500 Internal Server Error`

pschroedl:
Discovered a couple of upstream fixes that hadn't been committed - we weren't referencing the fork(s) with necessary dependency changes.  Will wait for a successful image build check to complete before merging.

runner/docker/Dockerfile.live-base-comfyui
victorges: Yeah! I think a `WORKDIR` on the `custo_nodes` folder and then setting up all the custom nodes in a contiguous "section" would look good :). Then also don't even need the `cd` before each clone

runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -32,8 +38,23 @@ RUN pip install git+https://github.com/yondonfu/comfystream.git
...
+
+# Install ComfyUI-SAM2-Realtime (https://github.com/pschroedl/ComfyUI-SAM2-Realtime.git)
+RUN cd /comfyui/custom_nodes && \
+    git clone https://github.com/pschroedl/ComfyUI-SAM2-Realtime.git && \
+    cd ComfyUI-SAM2-Realtime && \
+    pip install -r requirements.txt
+
+# Install ComfyUI-StreamDiffusion (https://github.com/pschroedl/ComfyUI-StreamDiffusion.git)
+RUN cd /comfyui/custom_nodes && \
+    git clone https://github.com/pschroedl/ComfyUI-StreamDiffusion.git && \
```
victorges, line 51: But another thing, which can be done separately if you prefer, is that we're not pinning the version of any of these nodes. So it could happen that the owners do a breaking change in their repo and it breaks our setup without us changing anything.

I would consider doing [something like this](https://github.com/livepeer/ai-worker/blob/ca5d650f9ebb60ee4ff8525b4076d19ba93fd169/runner/docker/Dockerfile.live-base-liveportrait#L44-L47), like cloning the repo, cd'ing inside and then checking out a specific commit. For the liveportrait one I just picked up the hash of the latest commit, which you can check easily with a command like this: `git ls-remote https://github.com/warmshao/FasterLivePortrait.git HEAD` (just learned this from GPT lol pretty cool)

pschroedl: That is a great idea - will update.

This will also allow us to validate changes to the more nodes easily without changing the version on the image that is in production.

eliteprox (approved):
LGTM! Tested with SAM2 and DepthAnything workflow configs (see attached)

[SAM2-Realtime-Example.json](https://github.com/user-attachments/files/18043947/SAM2-Realtime-Example.json)
[depth.json](https://github.com/user-attachments/files/18043953/depth.json)

PR #318, opened 2024-12-06
change github build action to run on oxford for live-pipeline builds

pschroedl:
docker builds are failing due to running out of space on the ubuntu24-04 runner, switching to oxford to un-break build for PR https://github.com/livepeer/ai-worker/pull/316

### DIFF
--- a/.github/workflows/ai-runner-live-pipelines-docker.yaml
+++ b/.github/workflows/ai-runner-live-pipelines-docker.yaml
@@ -26,7 +26,7 @@ jobs:
     name: Build common live base image
     permissions:
       pull-requests: read
-    runs-on: ubuntu-20.04
+    runs-on: oxford
     steps:
       - name: Check out code
         uses: actions/checkout@v4.1.1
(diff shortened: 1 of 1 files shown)


j0sh (approved):
I don't know what oxford is but if this unblocks things for now LGTM

We really should clear up some disk space on the runner though - cc @hjpotter92 @pwilczynskiclearcode

hjpotter92:
> I don't know what oxford is but if this unblocks things for now LGTM
> 
> We really should clear up some disk space on the runner though - cc @hjpotter92 @pwilczynskiclearcode

`ubuntu-20` tag refers to github's provided runner, which are limited to 14gb disk space. just switching the tag to `self-hosted` would have been fine. oxford limits it to just one runner, whereas we have 4 self-hosted runners (each with ~300GB capacity).

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

PR #319, opened 2024-12-09
chore(runner): remove duplicate pynvml dep

rickstaa:
This pull request removes the duplicate pynvml dependency.


runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -35,14 +35,14 @@ RUN cd /comfyui/custom_nodes && \
     cd ComfyUI-Misc-Effects && \
     git checkout c6b360c78611134c3723388170475eb4898ff6b7
 
-RUN pip install torch==2.5.1 torchvision torchaudio tqdm nvidia-ml-py==12.560.30
```
rickstaa, line 38: @leszko The only scenario where hardcoding this in the Dockerfile seems necessary is when a dependency chain installs an incorrect version of `nvidia-ml-py` before the `requirements.txt` file is processed. If this is the case we should remove it from the requirements so that we don't have two definitions that can conflict. Can you check that it works without hardcoding it in the docker?

rickstaa:
@victorges this was a part duplicate of https://github.com/livepeer/ai-worker/pull/323. Maybe we can add the comment so engineers know that the library is already included under a different name.

victorges:
@rickstaa hmmm that's super weird then, do you have an idea why were we getting errors when importing pynvml on some images like liveportrait?

mjh1:
I get the same error with streamdiffusion too (running on tensordoc) e.g. `docker run livepeer/ai-runner:live-app-streamdiffusion-675f99a`
Screenshot 2024-12-09 at 16 55 19

leszko:
> I get the same error with streamdiffusion too (running on tensordoc) e.g. `docker run livepeer/ai-runner:live-app-streamdiffusion-675f99a` Screenshot 2024-12-09 at 16 55 19

Yeah, then I suggest not merging this PR, because it will break our environments.

We need to research and fix these dependencies, but it's not the highest priority right now

leszko (changes requested):
As stated by @mjh1 this breaks the streamdiffusion + TRT runtime.

PR #324, opened 2024-12-09
runner/live/trickle: Fix publisher/subscribe lifecycle

victorges:
We weren't properly closing the sessions before so there was a warning from aiohttp

### DIFF
--- a/runner/app/live/trickle/media.py
+++ b/runner/app/live/trickle/media.py
@@ -185,4 +185,5 @@ async def run_publish(publish_url: str, image_generator):
         logging.error(f"postprocess got error {e}", e)
         raise e
     finally:
-        await publisher.close()
+        if publisher:
+            await publisher.close()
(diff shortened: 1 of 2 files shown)


runner/app/live/trickle/media.py
```diff
@@ -130,55 +130,55 @@ def feed_ffmpeg(ffmpeg_fd, image_generator):
...
-        segment_thread.start()
-        ffmpeg_feeder.start()
-        logging.debug("run_publish: ffmpeg feeder and segmenter threads started")
-
-        def joins():
-            segment_thread.join()
-            ffmpeg_feeder.join()
-        await asyncio.to_thread(joins)
-        logging.info("run_publish complete")
+        async with TricklePublisher(url=publish_url, mime_type="video/mp2t") as publisher:
```
j0sh, line 133: I believe `publisher.close` is already called as part of the `finally` ; is that still necessary here?

victorges: Fixed!

PR #327, opened 2024-12-09
Init prompt from params

pschroedl:
use prompt from params if present when the comfyui pipeline starts, add logging when prompt is set


eliteprox (approved):
LGTM! This allows us to begin a stream with specified workflow json config. Tested with SAM2

eliteprox:
Pinned to streamdiffusion fork to resolve hugging_face version issue
https://github.com/livepeer/ai-worker/pull/327/files#diff-f31feccb2794231dcf1486973c09d7c443f7a49150113cedd78039b5604919d0R28

PR #328, opened 2024-12-09
(chore) remove cleanup-runner for self-hosted runners

eliteprox:
This change resolves an issue with docker build workflows that use a self-hosted runner


pschroedl (approved):
LGTM - Lets Gooo!

PR #330, opened 2024-12-09
ci(runner): change docker driver for self-hosted runners

rickstaa:
This pull request changes the docker-driver to `docker-container` for the pipline build jobs that use `self-hosted` runners.


rickstaa:
The Docker Hub auth problem was caused by the mac runner so this is no longer needed.

PR #333, opened 2024-12-09
ci(runner): remove cancel in context comment

rickstaa:
This pull request removes the cancel in context comment since it is the desired behavoir.


rickstaa:
Since I think `false` is the default we can also merge https://github.com/livepeer/ai-worker/pull/334.

victorges (approved):
LGTM

PR #334, opened 2024-12-09
ci(runner): remove cancel in progress option from pipline docker CI

rickstaa:
This pull request removes the `cancel-in-progress` workflow action option since it is the desired behavoir.


rickstaa:
Closed because we choose https://github.com/livepeer/ai-worker/pull/333.

PR #343, opened 2024-12-11
feat: Add florence 2 to comfyui container

ad-astra-video:
Add Florence 2 to comfyui container. 

Docker container builds locally and temporarily pushed to `adastravideo/ai-runner:live-app-comfyui` Tested locally through comfystream and runs only on first frame to select a point and when inputs change to the Florence2run node for task, text_input, annotation_color, and output_mask_select.


leszko (approved):
Nice work Brad! 🏅 👏

PR #344, opened 2024-12-11
bump commit hash for streamdiffusion node, update dl_checkpoints.sh

pschroedl:
Update streamdiffusion node to include improvements in :
https://github.com/pschroedl/ComfyUI-StreamDiffusion/pull/5

Updates DL_checkpoints to pull .safetensor checkpoint files as expected by new loading mechanism.

This should probably be fixed upstream in the node to use proper HF caching. Attempts were made to symlink the existing huggingface-cli downloaded models but that approach is not straightforward and this is a workaround that should be revisited.


leszko (approved):
Nice, good work @pschroedl I see the perf is improved from 4FPS to 10-11 FPS. Nice!

PR #353, opened 2024-12-11
runner/pipelines: Accept ComfyUI prompt as a string

victorges:
This improves the ComfyUI params validation by:
- Creating a pydantic model to validate the provided input, similar to other pipelines
- Accepting the prompt as a string as well, which seems to be what the showcase app is sending for now. We make sure to parse it and send it as a JSON to comfystream, which is what they expect


eliteprox (approved):
Changes LGTM! Tested in staging and with e2e tests

leszko (approved):

PR #357, opened 2024-12-12
dl_checkpoints.sh --tensorrt - exit if previous build is still running

pwilczynskiclearcode:


hjpotter92 (approved):

runner/dl_checkpoints.sh
```diff
@@ -108,12 +108,17 @@ function download_stable_diffusion_checkpoints() {
 function build_tensorrt_models() {
     download_live_models
 
+    if [[ "$( docker ps -a -q --filter="label=TensorRT-engines" )" ]]; then
```
victorges, line 111: Are the builds asynchronous now?:

pwilczynskiclearcode: No they are synchronous... we run it from ansible and experience various issues. It looks like some of these containers are running forever without exiting correctly. Checking directly the logs from containers show no progress. Then killing the `dl_checkpoints.sh` by ansible timeout makes the containers to leak.

I think the proper solution is to build tensorrt once, store on S3 and then just pull it on the machine or development env. But that's sth to be planned for the next weeks.

PR #358, opened 2024-12-12
dl_checkpoints.sh new --batch flag to install only sam2-hiera-large model

pwilczynskiclearcode:
Adjusting the script so it work also for AI Batch orchestrators

### DIFF
--- a/runner/dl_checkpoints.sh
+++ b/runner/dl_checkpoints.sh
@@ -15,6 +15,9 @@ function display_help() {
     echo "Options:"
     echo "  --beta  Download beta models."
     echo "  --restricted  Download models with a restrictive license."
+    echo "  --live  Download models only for the livestreaming pipelines."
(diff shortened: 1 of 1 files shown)


runner/dl_checkpoints.sh
```diff
@@ -15,6 +15,9 @@ function display_help() {
     echo "Options:"
     echo "  --beta  Download beta models."
     echo "  --restricted  Download models with a restrictive license."
+    echo "  --live  Download models only for the livestreaming pipelines."
+    echo "  --tensorrt  Download livestreaming models and build tensorrt models."
+    echo "  --batch  Download all models for batch processing."
```
pwilczynskiclearcode, line 20: I'm not sure what terminology we use. I think "batch" is fine

leszko (approved):
I think we call it `AI Jobs`, but batch is ok as well. No strong feelings about it (CC @rickstaa if you have any opinion).

PR #367, opened 2024-12-13
feat(runner): add transformers pipeline logic

rickstaa:
This pull request adds the initial transformers pipeline which makes use of the pipeline abstraction of the transformers package to call any huggingface transformers pipeline.


runner/app/routes/transformers.py
```diff
@@ -0,0 +1,160 @@
...
+    torch_device = get_torch_device()
+
+    # Initialize the pipeline with the specified task and model ID.
+    pipeline_kwargs = {}
+    if request.task:
+        pipeline_kwargs["task"] = request.task
+    if request.model_name:
+        pipeline_kwargs["model"] = request.model_name
+    try:
+        pipe = pipeline(device=torch_device, **pipeline_kwargs)
```
rickstaa, line 138: I'm haven't checked whether models are removed from the GPU afterward.

RUFFY-369: @rickstaa I think either we can perform cache cleaning with `empty_cache()` or use pipeline in context manager: `with pipeline(...)`

PR #375, opened 2024-12-18
dl_checkpoints.sh - restore docker pull for mutable docker tags

pwilczynskiclearcode:


leszko (approved):

runner/dl_checkpoints.sh
pwilczynskiclearcode, line 1: `docker image tag` is a temporary solution. The proper one is to make [var livePipelineToImage](https://github.com/livepeer/ai-worker/blob/main/worker/docker.go#L62-L68) configurable via envs/CLI/ai-models.json

PR #387, opened 2024-12-28
feat: better LLM response format

kyriediculous:
- Improve LLM responses (openAI compatible format for non-streaming responses)
- Use uuid for vLLM request generation


kyriediculous:
I don't know why exactly but I keep running into differently generated bindings than you guys. Might be a versioning thing.

runner/app/pipelines/llm.py
```diff
@@ -219,15 +221,22 @@ async def generate(
...
-                    }
+                    yield LLMResponse(
+                        choices=[
+                            LLMChoice(
+                                delta=LLMMessage(
+                                    role="assistant",
+                                    content=delta
+                                ),
+                                index=0,
+                                finish_reason=None
```
ad-astra-video, line 232: `None` here causes validation error:
```
2024-12-29 00:23:16,509 - app.pipelines.llm - ERROR - Error generating response: 1 validation error for LLMChoice
finish_reason
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
```
```suggestion
                                finish_reason="none"
```

ad-astra-video:
Reviewed the updates, some comments below:

- stream_generator needs to be updated to process the `LLMResponse` returned now. Below is how I did it to test end to end. You may want to do differently, just wanted to provide head start.
```
 async def stream_generator(generator):
     try:
         async for chunk in generator:
-            if isinstance(chunk, dict):
-                if "choices" in chunk:
+            if isinstance(chunk, LLMResponse):
+                if len(chunk.choices) > 0:
                     # Regular streaming chunk or final chunk
-                    yield f"data: {json.dumps(chunk)}\n\n"
-                    if chunk["choices"][0].get("finish_reason") == "stop":
+                    yield f"data: {chunk.model_dump_json()}\n\n"
+                    if chunk.choices[0].finish_reason == "stop":
                         break
```

- The `defer cancel()` in LLM function of worker.go needs to be removed, it returns the container immediately for streamed responses.  Was not able to test this until testing through go-livepeer managed containers.
```
+++ b/worker/worker.go
@@ -397,7 +397,7 @@ func (w *Worker) AudioToText(ctx context.Context, req GenAudioToTextMultipartReq
 func (w *Worker) LLM(ctx context.Context, req GenLLMJSONRequestBody) (interface{}, error) {
        isStreaming := req.Stream != nil && *req.Stream
        ctx, cancel := context.WithCancel(ctx)
-       defer cancel()
+
        c, err := w.borrowContainer(ctx, "llm", *req.Model)
        if err != nil {
                return nil, err
```
- The total_tokens seem to be double counted, maybe because the tokenizer is counting the space as a token? Is this expected behavior? note: this was not changed on this one and does not cause payment issues but wanted to confirm.

- Do you still want to use `LlmStreamChunk` to send the response through go-livepeer?  If want to change it to the LLMResponse will need to update go-livepeer but should be quick update. Example response chunk from streaming:
```
data: {"chunk":"{\"choices\":[{\"delta\":{\"role\":\"assistant\",\"content\":\"\"},\"index\":0,\"finish_reason\":\"stop\"}],\"tokens_used\":552,\"id\":\"chatcmpl-d2dd5298-8330-48db-8370-901dea8e5a12\",\"model\":\"meta-llama/Meta-Llama-3.1-8B-Instruct\",\"created\":1735437990}","tokens_used":552}
```

ad-astra-video:
@kyriediculous can you comment on these items above:

> handleStreamingResponse also needs update, example I did to test end to end

> The defer cancel() in LLM function of worker.go needs to be removed, it returns the container immediately for streamed responses. Was not able to test this until testing through go-livepeer managed containers.

> Do you still want to use LlmStreamChunk to send the response through go-livepeer?

ad-astra-video (commented):
Couple additional updates for change in response.

runner/app/routes/utils.py
```diff
@@ -77,6 +77,16 @@ class LLMMessage(BaseModel):
     content: str
 
 
+class LLMChoice(BaseModel):
+    delta: LLMMessage
+    index: int
+    finish_reason: str = ""
```
victorges, line 83: Is this default by design? AFAIK It will make all the `LLMChoice` objects in responses to contain an empty `finish_reason` field, instead of omitting the field when it's not a final chunk or sth

victorges (approved):
Code LGTM, but I haven't checked the OpenAPI schema to make sure the changes make it compatible. Only reviewed the changes here.

runner/app/routes/llm.py
```diff
@@ -71,24 +70,31 @@ async def llm(
...
-                id=last_chunk.get("id", ""),
-                model=last_chunk.get("model", pipeline.model_id),
-                created=last_chunk.get("created", int(time.time()))
-            )
+            if last_chunk:
+                # Return the final response with accumulated text
+                return LLMResponse(
+                    choices=[
+                        LLMChoice(
+                            delta=LLMMessage(
```
0xspeedybird, line 83: If I am not mistaken, this creates an invalid response.  The delta field is only used in streaming responses.  The LLMMessage should be put in a message field, not delta.

runner/app/routes/utils.py
```diff
@@ -77,6 +77,16 @@ class LLMMessage(BaseModel):
     content: str
 
 
+class LLMChoice(BaseModel):
+    delta: LLMMessage
```
0xspeedybird, line 81: Per my comment [above](https://github.com/livepeer/ai-worker/pull/387/files#r1909763218), this structure only is valid for streaming responses (if sticking to the openai spec).  That means a message field is needed on this class and is used for non-streaming responses.

runner/app/routes/utils.py
kyriediculous: Here's the generated go binding for the type, which would omit it when empty. It seems counterintuitive but it works.

```
type LLMChoice struct {
	Delta        LLMMessage `json:"delta"`
	FinishReason *string    `json:"finish_reason,omitempty"`
	Index        int        `json:"index"`
}
```

kyriediculous: It does return `finish_reason: ""` in the response however

runner/app/routes/llm.py
ad-astra-video: message field was added. example response below:

```
{
    "choices": [{
            "finish_reason": "stop",
            "index": 0,
            "message": {
                "content": "\u003c|start_header_id|\u003eassistant\u003c|end_header_id|\u003e\n\nBeep boop, I'm glad you asked. Here's a story about a robot named Zeta.\n\nIn the year 2157, in a world where robots had long surpassed human intelligence, I, Zeta, was created in a top-secret laboratory beneath the bustling city of New Tokyo.",
                "role": "assistant"
            }
        }
    ],
    "created": 1736833981,
    "id": "chatcmpl-ee2b50ca-c20c-4cb0-8e36-bde46d1acc90",
    "model": "meta-llama/Meta-Llama-3.1-8B-Instruct",
    "tokens_used": {
        "completion_tokens": 128,
        "prompt_tokens": 40,
        "total_tokens": 168
    }
}

```

runner/app/routes/utils.py
ad-astra-video: message field was added. example response below:

```
{
    "choices": [{
            "finish_reason": "stop",
            "index": 0,
            "message": {
                "content": "\u003c|start_header_id|\u003eassistant\u003c|end_header_id|\u003e\n\nBeep boop, I'm glad you asked. Here's a story about a robot named Zeta.\n\nIn the year 2157, in a world where robots had long surpassed human intelligence, I, Zeta, was created in a top-secret laboratory beneath the bustling city of New Tokyo.",
                "role": "assistant"
            }
        }
    ],
    "created": 1736833981,
    "id": "chatcmpl-ee2b50ca-c20c-4cb0-8e36-bde46d1acc90",
    "model": "meta-llama/Meta-Llama-3.1-8B-Instruct",
    "tokens_used": {
        "completion_tokens": 128,
        "prompt_tokens": 40,
        "total_tokens": 168
    }
}
```

ad-astra-video:
Approved! Will fix openapi gen in separate PR.

PR #389, opened 2025-01-03
feat: update A2T audio conversion

ad-astra-video:
PR improves file handling of `audio-to-text` pipeline by removing call to ffmpeg to process the audio and uses pyav to convert to raw audio.

Cloud SPE test file fails in my testing sometimes with current conversion implementation using "mp3" container and format. The processing error was coming from the internal conversion to raw audio in transformers (links below). Switching to processing the audio file using `pyav` allows the Cloud SPE file to process correctly.

Transformers calling ffmpeg binary from `preprocess` function.
https://github.com/huggingface/transformers/blob/47c29ccfaf56947d845971a439cbe75a764b63d7/src/transformers/pipelines/automatic_speech_recognition.py#L353
https://github.com/huggingface/transformers/blob/47c29ccfaf56947d845971a439cbe75a764b63d7/src/transformers/pipelines/audio_utils.py#L10

Some marginal speed improvements:
Cloud SPE test file (3s clip)
965ms - bytes of container sent to model
895ms - np ndarray sent to model

Another test file (3m 22s)
5.6s - bytes of container sent to model
5.5s - np ndarray sent to model

### DIFF
--- a/runner/app/pipelines/audio_to_text.py
+++ b/runner/app/pipelines/audio_to_text.py
@@ -12,6 +12,7 @@ from app.pipelines.base import Pipeline
 from app.pipelines.utils import get_model_dir, get_torch_device
 from app.pipelines.utils.audio import AudioConverter
 from app.utils.errors import InferenceError
+import numpy as np
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


runner/app/pipelines/utils/audio.py
```diff
@@ -19,7 +19,46 @@ class AudioConverter:
...
+        finally:
+            input_container.close()
+
+        return audio_array
+    
+    @staticmethod
+    def convert(input_bytes: bytes, output_extension: str, output_codec: str) -> bytes:
         """Converts an audio file to a different format.
 
         Args:
```
eliteprox, line 64: We can remove the convert method now, it's no longer used

eliteprox (approved):
LGTM! Tested with MP4 and regression MP3 and FLAC. Left a few nits for cleanup

runner/app/pipelines/utils/audio.py
ad-astra-video: Not used right now but I felt was worth keeping since we are still building out the feature.  Having a way to request a different audio format in return may be worth keeping around.

eliteprox: Sounds good to me

runner/app/pipelines/audio_to_text.py
```diff
@@ -120,13 +121,11 @@ def __init__(self, model_id: str):
...
-        if (
-            os.path.splitext(audio.filename)[1].lower().lstrip(".")
-            in INCOMPATIBLE_EXTENSIONS
-        ):
-            audioBytes = self._audio_converter.convert(audioBytes, "mp3")
+        #re-encode audio to match pre-processing done in transformers.
+        # pipeline accepts np.ndarray and does not convert it again. String file path and bytes are converted to np.ndarray in the pipeline.
+        #https://github.com/huggingface/transformers/blob/47c29ccfaf56947d845971a439cbe75a764b63d7/src/transformers/pipelines/automatic_speech_recognition.py#L353
+        #https://github.com/huggingface/transformers/blob/47c29ccfaf56947d845971a439cbe75a764b63d7/src/transformers/pipelines/audio_utils.py#L10
+        audio_array = self._audio_converter.to_ndarray(audioBytes)
```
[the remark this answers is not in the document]
ad-astra-video: Done

ad-astra-video:
I made updates for small clean up and removed import not needed

PR #394, opened 2025-01-08
Add support for LTX-Video model in ImageToVideo Pipeline

RUFFY-369:
# What does this PR do?

This PR adds support for the `LTX-Video` model in the `ImageToVideoPipeline`. It introduces a mechanism to dynamically load either the `LTXImageToVideoPipeline` or `StableVideoDiffusionPipeline`, based on the provided `modelID`. This change enhances the flexibility of the `ImageToVideoPipeline`, allowing it to handle both `LTX-Video` and `StableVideoDiffusion` models and with addition of `LTXImageToVideoPipeline`, the video generation can now be prompt text guided.

After building the Docker image, the setup has been tested by running it locally on a Uvicorn server.

cc @rickstaa

### DIFF
--- a/runner/app/pipelines/image_to_video.py
+++ b/runner/app/pipelines/image_to_video.py
@@ -6,7 +6,7 @@ from typing import Any, Dict, List, Optional, Tuple, Type
 
 import PIL
 import torch
-from diffusers import LTXImageToVideoPipeline, StableVideoDiffusionPipeline
+from diffusers import DiffusionPipeline, LTXImageToVideoPipeline, StableVideoDiffusionPipeline
 from huggingface_hub import file_download
 from PIL import ImageFile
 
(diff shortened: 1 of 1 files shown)


runner/app/pipelines/image_to_video.py
```diff
@@ -41,7 +43,28 @@ def __init__(self, model_id: str):
...
+                logger.info("Loading LTXImageToVideoPipeline for model_id: %s", model_id)
+                self.pipeline_name = "LTXImageToVideoPipeline"
+                self.ldm = LTXImageToVideoPipeline.from_pretrained(model_id, **kwargs)
+            else:
+                logger.info("Loading StableVideoDiffusionPipeline for model_id: %s", model_id)
+                self.pipeline_name = "StableVideoDiffusionPipeline"
+                self.ldm = StableVideoDiffusionPipeline.from_pretrained(model_id, **kwargs)
+        except Exception as loading_error:
+            logger.error("Failed to load %s : %s." %(self.pipeline_name,loading_error))
+            # Trying to load the LTXImageToVideoPipeline if the StableVideoDiffusionPipeline fails to load and there is a chance that model name doesn't match the if condition for LTX-Video
```
ad-astra-video, line 57: I think we should exclude retrying the LTXImageToVideoPipeline load again here. 

Would switching to use `DiffusionPipeline` make the loading generic? We are only passing model_id and kwargs so seems like most of it is setup to use generic loading if its possible

runner/app/pipelines/image_to_video.py
```diff
@@ -113,6 +136,14 @@ def __call__(
         seed = kwargs.pop("seed", None)
         safety_check = kwargs.pop("safety_check", True)
 
+        if self.pipeline_name == "LTXImageToVideoPipeline":
```
ad-astra-video, line 139: Can we do something more generic that looks at the pipeline class args and del the kwargs keys if not present in the pipeline? If a heavy lift we can do in separate PR.

runner/app/pipelines/image_to_video.py
```diff
@@ -41,7 +43,28 @@ def __init__(self, model_id: str):
...
+                self.pipeline_name = "LTXImageToVideoPipeline"
+                self.ldm = LTXImageToVideoPipeline.from_pretrained(model_id, **kwargs)
+            except Exception as loading_error:
+                logger.error("Failed to load both LTXImageToVideoPipeline and StableVideoDiffusionPipeline: %s. Please ensure the model ID is compatible.", loading_error)
+                raise loading_error
+            
+
         self.ldm.to(get_torch_device())
 
         sfast_enabled = os.getenv("SFAST", "").strip().lower() == "true"
```
ad-astra-video, line 70: Need to confirm if SFAST works on LTXImageToVideoPipeline. If it does not, only use it if `StableVideoDiffusionPipeline` is loaded.

worker/multipart.go
```diff
@@ -112,6 +112,16 @@ func NewImageToVideoMultipartWriter(w io.Writer, req GenImageToVideoMultipartReq
 		return nil, fmt.Errorf("failed to copy image to multipart request imageBytes=%v copiedBytes=%v", imageSize, copied)
 	}
 
+	if req.Prompt != nil {
+		if err := mw.WriteField("prompt", req.Prompt); err != nil {
+			return nil, err
```
ad-astra-video, line 117:
```suggestion
		if err := mw.WriteField("prompt", *req.Prompt); err != nil {
```

worker/multipart.go
```diff
@@ -112,6 +112,16 @@ func NewImageToVideoMultipartWriter(w io.Writer, req GenImageToVideoMultipartReq
 		return nil, fmt.Errorf("failed to copy image to multipart request imageBytes=%v copiedBytes=%v", imageSize, copied)
 	}
 
+	if req.Prompt != nil {
+		if err := mw.WriteField("prompt", req.Prompt); err != nil {
```
ad-astra-video, line 116:
```suggestion
		if err := mw.WriteField("prompt", *req.Prompt); err != nil {
```

ad-astra-video (commented):
PR looks good, have couple comments below and left suggestions/questions in the files:

- LTXVideo requires diffusers 0.32.0 right? Suggest we update to diffusers 0.32.1 with this PR.
- If fix the multipart.go suggested change I can build and test tomorrow

runner/app/pipelines/image_to_video.py
rickstaa: Good point would love to make all pipelines more generic. We can do this in seperate PR.

runner/app/pipelines/image_to_video.py
ad-astra-video: confirmed this works with local build

runner/app/pipelines/image_to_video.py
ad-astra-video: Thanks for confirming

ad-astra-video:
@RUFFY-369 I rebased this and removed the parts moved to livepeer/go-livepeer 

https://github.com/livepeer/ai-runner/pull/481

PR #397, opened 2025-01-22
dl_checkpoints.sh - docker run as a user not the root

pwilczynskiclearcode:
By default `docker run` runs as `root` and that mixes `/models/` dirs and files ownership


rickstaa:
Thanks!

victorges:
Thanks!! [2]

PR #400, opened 2025-01-24
Audio passthrough

j0sh:
Passes through any audio in the input to the output.

Notes:
- Had to switch to PyAV rather than ffmpeg for better access to timestamps. This resulted in a bunch of cleanup for supporting code that is no longer needed (but also a lot of new additions)
- We also had to bump up the PyAV version in the Dockerfiles since it was relying on an ancient version of libavcodec/libavformat; hopefully this is not a problem for whatever other dependencies are using PyAV. We need a recent version to able to support Opus inputs.
- If no audio is present in the input, no audio will be in the output.
- Likewise, if no _video_ if present in the input, then no video in the output ... we will just pass through the audio 🤷 
- AAC is the output codec. For WebRTC delivery, the audio will need to be transcoded to something like Opus or G711.
- This downmixes to mono regardless of the input layout / channels (since I suspect that most inference will output mono) but it is easy to change that
- There is no audio pipeline wired in - not even a noop pipeline. For any audio frames we pop from the input queue, we push directly into the output queue.
- The output queue is currently unbounded. This is a pre-existing issue and not a new one in this PR; we can fix it later, however we might want to prioritize audio rather than video when dropping.
- This PR does not affect event metrics (or at least, it tries not to). The existing event metrics are video-only. We probably should define and track some other metrics for audio, but that can be done later.
- Only tested with trickle protocol / go-livepeer e2e


runner/app/live/streamer/process.py
```diff
@@ -136,13 +137,20 @@ def report_error(error_msg: str):
...
                     continue
 
                 try:
-                    output_image = pipeline.process_frame(input_image)
-                    self.output_queue.put(output_image)
+                    if isinstance(input_frame, VideoFrame):
+                        output_image = pipeline.process_frame(input_frame.image)
+                        output_frame = VideoOutput(input_frame.replace_image(output_image))
+                        self.output_queue.put(output_frame)
+                    elif isinstance(input_frame, AudioFrame):
```
victorges, line 149: neat

runner/app/live/trickle/frame.py
```diff
@@ -0,0 +1,81 @@
...
+    @property
+    def timestamp(self):
+        return self.frame.timestamp
+
+    @property
+    def time_base(self):
+        return self.frame.time_base
+
+class AudioOutput(OutputFrame):
+    frames: List[AudioFrame]
```
varshith15, line 79: whys is this a list any specific reason? comfystream does the list handling, so this might not be needed?

j0sh: Yes - and Comfy should output the list when it's done processing, to minimize IPC and pressure on the queue. Since we know exactly what needs to be sent, send it all at once.

varshith15: but comfystream has a list already, you call a function to get the next frame https://github.com/varshith15/comfystream/blob/af132bedd665027fa02c20601b98e0c229e54f51/server/pipeline.py#L74

we want it to be frame by frame in audio as well, otherwise we gotta maintain all the frame instances that were sent into the pipeline to assign processed data back into the the frames(ref line no 150 in runner/app/live/streamer/process.py)

it's just simpler to have it as a single frame because a list is maintained inside the comfystream already

runner/app/live/streamer/protocol/trickle.py
```diff
@@ -6,30 +6,30 @@
...
 
 class TrickleProtocol(StreamProtocol):
     def __init__(self, subscribe_url: str, publish_url: str, control_url: Optional[str] = None, events_url: Optional[str] = None):
         self.subscribe_url = subscribe_url
         self.publish_url = publish_url
         self.control_url = control_url
         self.events_url = events_url
-        self.subscribe_queue = queue.Queue[bytearray]()
-        self.publish_queue = queue.Queue[bytearray]()
+        self.subscribe_queue = queue.Queue[InputFrame]()
```
varshith15, line 19: why not use asyncio queues?

j0sh: Not sure; original code was not asyncio and this PR doesn't change anything there

runner/docker/Dockerfile.live-base
```diff
@@ -37,26 +37,26 @@ RUN git clone -b n12.2.72.0 --depth 1 https://git.videolan.org/git/ffmpeg/nv-cod
...
-RUN git clone --branch n7.0.2 --depth 1 https://github.com/FFmpeg/FFmpeg.git ffmpeg
+RUN git clone --branch n7.1 --depth 1 https://github.com/FFmpeg/FFmpeg.git ffmpeg
 
 # Build FFmpeg with static linking and the desired hardware-accelerated features
 RUN cd ffmpeg && \
     ./configure --prefix=/compiled \
-                --pkg-config-flags="--static" \
                 --extra-cflags="-I/ffmpeg_build/include -I/usr/local/cuda/include" \
                 --extra-ldflags="-L/ffmpeg_build/lib -L/usr/local/cuda/lib64" \
-                --extra-libs="-lpthread -lm" \
```
emranemran, line 48: is removing libpthread expected here? I'm assuming various parts of ffmpeg rely on it to run efficiently?

j0sh: IIRC it was there to support static linking; I think there was a bug with the pkg-config declaration since that should also be pulling in pthreads.

It is still pulled in dynamically -

```
$ docker run -it --entrypoint /bin/bash livepeer/ai-runner:live-app-noop
root@733b39a972a0:/app# ldd /usr/local/bin/ffmpeg 
        linux-vdso.so.1 (0x00007ffd429f7000)
        libavdevice.so.61 => /usr/local/lib/libavdevice.so.61 (0x00007ff488af1000)
        libavfilter.so.10 => /usr/local/lib/libavfilter.so.10 (0x00007ff488660000)
        libavformat.so.61 => /usr/local/lib/libavformat.so.61 (0x00007ff4883cf000)
        libavcodec.so.61 => /usr/local/lib/libavcodec.so.61 (0x00007ff487058000)
        libpostproc.so.58 => /usr/local/lib/libpostproc.so.58 (0x00007ff487045000)
        libswresample.so.5 => /usr/local/lib/libswresample.so.5 (0x00007ff487024000)
        libswscale.so.8 => /usr/local/lib/libswscale.so.8 (0x00007ff486f8e000)
        libavutil.so.59 => /usr/local/lib/libavutil.so.59 (0x00007ff485e89000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff485da2000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff485b7a000)
        libxcb.so.1 => /lib/x86_64-linux-gnu/libxcb.so.1 (0x00007ff485b50000)
        libnppig.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppig.so.12 (0x00007ff483400000)
        libnppicc.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppicc.so.12 (0x00007ff482a00000)
        libnppidei.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppidei.so.12 (0x00007ff481c00000)
        libnppif.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppif.so.12 (0x00007ff47be00000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007ff485b32000)
        libbz2.so.1.0 => /lib/x86_64-linux-gnu/libbz2.so.1.0 (0x00007ff485b1f000)
        liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007ff4833d5000)
        libfdk-aac.so.2 => /lib/x86_64-linux-gnu/libfdk-aac.so.2 (0x00007ff4828cf000)
        libopus.so.0 => /lib/x86_64-linux-gnu/libopus.so.0 (0x00007ff483377000)
        libx264.so.163 => /lib/x86_64-linux-gnu/libx264.so.163 (0x00007ff47bb3f000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff488b6b000)
        libXau.so.6 => /lib/x86_64-linux-gnu/libXau.so.6 (0x00007ff483371000)
        libXdmcp.so.6 => /lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007ff483369000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff483364000)
        libnppc.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppc.so.12 (0x00007ff47b600000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007ff48335f000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ff48335a000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff48333a000)
        libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007ff483322000)
        libmd.so.0 => /lib/x86_64-linux-gnu/libmd.so.0 (0x00007ff483315000)
root@733b39a972a0:/app# 
```

runner/app/live/trickle/media.py
```diff
@@ -149,36 +93,47 @@ async def callback(pipe_file, pipe_name):
...
-        ffmpeg_feeder.start()
-        logging.debug("run_publish: ffmpeg feeder and segmenter threads started")
-
+        def sync_callback(pipe_file, pipe_name):
+            def do_schedule():
+                schedule_callback(callback(pipe_file, pipe_name), pipe_name)
+            loop.call_soon_threadsafe(do_schedule)
+
+        # hold tasks since `loop.create_task` is a weak reference that gets GC'd
+        # TODO use asyncio.TaskGroup once all pipelines are on Python 3.11+
```
emranemran, line 102: Can we not switch to python 3.11 right now? Is there a limitation? The logic here looks alright but it feels like using TaskGroup would simplify it a bit to avoid manual tracking of tasks. We can merge liek this for now and re-test but perhaps should consider revisiting before launch if we see issues? wdyt? 

cc @victorges as well for comments on this since I'm still relearning Pythonic concurrency.

runner/app/live/trickle/frame.py
emranemran: @varshith15 maybe you can handle this when you do the ai-runner integration and we leave as-is for now since first goal is to get passthru working?

emranemran (approved):

emranemran:
Left a few generic comments on things we may need follow-up on - but I think it looks good to merge to staging and get some testing going before spending too much time with local testing and iteration. wdyt?

runner/app/live/trickle/decoder.py
```diff
@@ -0,0 +1,106 @@
...
+                    if aframe.pts is None:
+                        continue
+
+                    avframe = InputFrame.from_av_audio(aframe)
+                    frame_callback(avframe)
+                    continue
+
+            elif video_stream and packet.stream == video_stream:
+                # Decode video frames
+                for frame in packet.decode():
```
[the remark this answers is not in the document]
j0sh: Software only for now. The old ffmpeg code was also software-only (there was logic for GPU but we never enabled it)

We can enable GPU here eventually but want to keep this like-for-like to avoid introducing new issues, IIRC we've seen [problems](https://github.com/livepeer/go-livepeer/blob/9508e13d3004bd03c6c16303c99412116683379c/core/ai_orchestrator.go#L1097-L1102) with transcoding + AI workloads

runner/app/live/trickle/media.py
j0sh: I believe there are dependency limitations in some pipelines that preclude us from using newer Python although I am not sure what the minimum version would be - @victorges might know better. Didn't want to introduce the risk of a Python upgrade and pipeline validation alongside this PR which is already invasive enough as-is

runner/app/live/trickle/frame.py
varshith15: cool, makes sense, can pick this up later then integrating comfystream package

PR #408, opened 2025-01-31
test docker for audio passthrough

j0sh:
#400 had the wrong repo; pushing the same code here to make sure docker builds


emranemran (approved):

j0sh:
CI is red because the Docker live pipelines are [failing to build](https://github.com/livepeer/ai-runner/actions/runs/13080738183/job/36503997033) and it looks like we have a chicken-and-egg problem: we do not push images to Dockerhub [if they came from a PR](https://github.com/livepeer/ai-runner/blob/a35bb9cffb1c11fb0059946a88d1eff722fb6b6b/.github/workflows/ai-runner-live-pipelines-docker.yaml#L71) (which makes sense if we just want to check that they build), but this creates a chicken-and-egg problem if any other images depend on that updated one ... which is exactly what happened here, since we updated the live base image but the pipelines are not picking up on that.

The solution might be to just merge anyway and hope for the best? Things seem to build fine locally.

Merged via #400

PR #411, opened 2025-02-06
Remove Manager custom node in speed up the startup

leszko:


pschroedl (approved):
Agreed - should shave off some startup time.  LGTM.

emranemran (approved):

PR #419, opened 2025-02-13
fix container monitoring by moving blocking pipeline call to asyncio thread

ad-astra-video:
This updates the pipeline calls to move to thread in event loop which fixes containers being killed if not responding to the /health checks the ai-worker does in go-livepeer.  

I have seen this on FLUX DEV requests and image-to-video pipeline requests.

It happens sometimes on SFAST initialization as well.  Should we consider moving that to the event loop or move all of the calls to the pipeline functions to the event loop?

See discord thread of report from Orchestrator running image-to-video: https://discord.com/channels/423160867534929930/1338172672152047739/1338586158321504426

cc @victorges @rickstaa

### DIFF
--- a/runner/app/routes/image_to_image.py
+++ b/runner/app/routes/image_to_image.py
@@ -1,3 +1,5 @@
+import asyncio
+from functools import partial
 import logging
 import os
 import random
@@ -182,19 +184,21 @@ async def image_to_image(
     has_nsfw_concept = []
     for seed in seeds:
         try:
-            imgs, nsfw_checks = pipeline(
-                prompt=prompt,
-                image=image,
-                strength=strength,
-                loras=loras,
-                guidance_scale=guidance_scale,
-                image_guidance_scale=image_guidance_scale,
-                negative_prompt=negative_prompt,
-                safety_check=safety_check,
-                seed=seed,
-                num_images_per_prompt=1,
-                num_inference_steps=num_inference_steps,
-            )
+            pipeline_call = partial(pipeline, 
+                                    prompt=prompt,
+                                    image=image,
+                                    strength=strength,
+                                    loras=loras,
+                                    guidance_scale=guidance_scale,
+                                    image_guidance_scale=image_guidance_scale,
+                                    negative_prompt=negative_prompt,
+                                    safety_check=safety_check,
+                                    seed=seed,
+                                    num_images_per_prompt=1,
+                                    num_inference_steps=num_inference_steps,
+                                    scheduler=scheduler
+                                )
+            imgs, nsfw_checks = await asyncio.to_thread(pipeline_call)
         except Exception as e:
             if isinstance(e, torch.cuda.OutOfMemoryError):
                 # TODO: Investigate why not all VRAM memory is cleared.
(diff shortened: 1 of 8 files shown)


runner/app/routes/upscale.py
```diff
@@ -126,13 +128,13 @@ async def upscale(
...
 
     try:
-        images, has_nsfw_concept = pipeline(
-            prompt=prompt,
-            image=image,
-            num_inference_steps=num_inference_steps,
-            safety_check=safety_check,
-            seed=seed,
-        )
+        pipeline_call = partial(pipeline,
```
rickstaa, line 131: @ad-astra-video This issue can also be resolved by removing the `async` keyword from the pipeline class. When FastAPI doesn't detect `async`, a separate thread for the route is spawned. The downside is that it can't handle multiple requests concurrently.  

The current async implementation hangs the application because it expects an `await`, but instead encounters a blocking operation during model inference. Your solution works but introduces potential issues if the worker doesn't properly manage capacity. This could lead to multiple inputs running through the same shared model instance, causing errors like:  

```bash
timestamp=2025-02-13 14:16:04 level=ERROR location=text_to_image.py:206:text_to_image request_id= stream_id= message=TextToImage pipeline error: Error during model execution: index 51 is out of bounds for dimension 0 with size 51
```

However, I noticed that @yondonfu originally used the `async` keyword as well, so I may be overlooking something. We could implement locks to prevent race conditions, but that adds complexity. If Yondon agrees, I think we should refactor the batch loops without `async`.

rickstaa: The `/health` endpoint can remain `sync` since we don’t anticipate concurrent calls from the worker. However, the hardware endpoints maybe should be made `async` by wrapping the hardware request logic with `asyncio.to_thread`. This ensures they can handle simultaneous requests from multiple Gateways in the future.

rickstaa (changes requested):
Let's wait what @yondonfu or @emranemran think.

runner/app/routes/image_to_image.py
```diff
@@ -182,19 +184,21 @@ async def image_to_image(
...
+                                    strength=strength,
+                                    loras=loras,
+                                    guidance_scale=guidance_scale,
+                                    image_guidance_scale=image_guidance_scale,
+                                    negative_prompt=negative_prompt,
+                                    safety_check=safety_check,
+                                    seed=seed,
+                                    num_images_per_prompt=1,
+                                    num_inference_steps=num_inference_steps,
+                                    scheduler=scheduler
```
rickstaa, line 199: If we go with this solution then scheduler should be defined.

runner/app/routes/upscale.py
rickstaa: This should work -> https://github.com/ad-astra-video/ai-worker/pull/2/files.

runner/app/routes/image_to_image.py
ad-astra-video: This was removed

runner/app/routes/upscale.py
ad-astra-video: merged

PR #423, opened 2025-02-13
Add a version metric

mjh1:
Add a prometheus metric to record the version of the code running, so that we can track this in our dashboards.

### DIFF
--- a/.github/workflows/ai-runner-live-pipelines-docker.yaml
+++ b/.github/workflows/ai-runner-live-pipelines-docker.yaml
@@ -153,7 +153,7 @@ jobs:
           file: docker/Dockerfile.live-app__PIPELINE__
           build-args: |
             PIPELINE=${{ matrix.pipeline }}
-            GIT_SHA=$(git rev-parse HEAD)
+            GIT_SHA=${{ (github.ref_type == 'tag' && github.ref_name) || (github.event.pull_request.head.sha || github.sha) }}
           cache-from: type=registry,ref=livepeerci/build:cache
           cache-to: type=registry,ref=livepeerci/build:cache,mode=max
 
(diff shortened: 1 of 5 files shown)


runner/app/main.py
```diff
@@ -7,10 +7,13 @@
 from fastapi.routing import APIRoute
 from app.utils.hardware import HardwareInfo
 from app.live.log import config_logging
+from prometheus_client import Counter, generate_latest, CONTENT_TYPE_LATEST
+from starlette.responses import Response
 
 config_logging()
 logger = logging.getLogger(__name__)
 
+VERSION = Counter('version', 'Runner version', ['app', 'version'])
```
hjpotter92, line 16: the other version metrics are `Gauge` type. i don't know if it'd conflict with `Counter` or not. https://github.com/livepeer/studio/blob/2bb3b0ab4bd015e3e561266072580192af25cca3/packages/api/src/index.ts#L129-L142

when querying, we simply use the promql expression:

```
sum by (app, version) version{}
```

.github/workflows/ai-runner-live-pipelines-docker.yaml
```diff
@@ -153,6 +153,7 @@ jobs:
           file: docker/Dockerfile.live-app__PIPELINE__
           build-args: |
             PIPELINE=${{ matrix.pipeline }}
+            GIT_SHA=$(git rev-parse HEAD)
```
hjpotter92, line 156: need to use github expression here, as it doesn't run bash commands here:
```yaml
GIT_SHA=${{ (github.ref_type == 'tag' && github.ref_name) || (github.event.pull_request.head.sha || github.sha) }}
```

hjpotter92 (approved):

PR #424, opened 2025-02-13
infer.py: Run a single pipeline process for all streams

victorges:
We want to improve cold start time for a stream so can't afford the isolation of
having one process per stream. Restructure infer.py so it is started with the
whole container and re-used across streams.

This was accomplished by removing the process-managemenet logic from the Streamer
and moving to an outer process guardian, then only create ephemeral streamers that use
the process guardian. 

The infer.py now exposes an API to be called to start the streamer as well which is
called from the runner API. The process is started immediately with the runner API
startup (from pipeline creation).

### DIFF
--- a/runner/app/live/streamer/process_guardian.py
+++ b/runner/app/live/streamer/process_guardian.py
@@ -26,7 +26,8 @@ class ProcessGuardian:
 
     async def start(self):
         # Start the pipeline process and initialize timing
-        self.process = PipelineProcess.start(self.pipeline, self.params)
+        # TODO: Fix contextual logging on request_id and stream_id
+        self.process = PipelineProcess.start(self.pipeline, self.params, "", "")
         if self.process is None:
             raise RuntimeError("Failed to start PipelineProcess")
         # Launch the monitor loop as a background task
@@ -159,7 +160,8 @@ class ProcessGuardian:
         # don't call the full start/stop methods since we only want to restart the process
         await self.process.stop()
 
-        self.process = PipelineProcess.start(self.pipeline, self.params)
+        # TODO: Fix contextual logging on request_id and stream_id
+        self.process = PipelineProcess.start(self.pipeline, self.params, "", "")
         self.status.inference_status.restart_count += 1
         self.status.inference_status.last_restart_time = time.time()
         self.status.inference_status.last_restart_logs = restart_logs
--- a/runner/app/live/streamer/process_guardian.py
+++ b/runner/app/live/streamer/process_guardian.py
@@ -30,8 +30,7 @@ class ProcessGuardian:
 
     async def start(self):
         # Start the pipeline process and initialize timing
-        # TODO: Fix contextual logging on request_id and stream_id
-        self.process = PipelineProcess.start(self.pipeline, self.params, "", "")
+        self.process = PipelineProcess.start(self.pipeline, self.params)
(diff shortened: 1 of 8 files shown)


runner/app/live/streamer/process_guardian.py
```diff
@@ -0,0 +1,276 @@
...
+        if not self.process:
+            raise RuntimeError("Process not running")
+        self.status = PipelineStatus(
+            pipeline=self.pipeline, start_time=self.status.start_time
+        )
+        self.monitoring_callback = monitoring_callback
+        self.process.reset_stream(request_id, stream_id)
+        await self.update_params(params)
+
+    def send_input(self, frame: InputFrame):
```
emranemran, line 67: what's the difference between ProcessGuardian handling input/output of frames vs the PipelineStreamer? I'm having a bit of confusion coming up with a mental model

runner/app/live/streamer/process_guardian.py
```diff
@@ -0,0 +1,276 @@
...
+        self.process.send_input(frame)
+
+    async def recv_output(self) -> OutputFrame | None:
+        if not self.process:
+            raise RuntimeError("Process not running")
+        output = await self.process.recv_output()
+
+        oss = self.status.inference_status
+        previous_output_time = max(oss.last_output_time or 0, self.process.start_time)
+        (oss.last_output_time, oss.fps) = calculate_rolling_fps(
```
emranemran, line 85: is rolling input/output fps logged anywhere? Is it the same as INPUT/OUTPUT FPS reported or is that the instantaneous measurement?

runner/app/live/streamer/process_guardian.py
```diff
@@ -0,0 +1,276 @@
...
+        time_since_last_input = current_time - last_input_time
+        if time_since_last_input > 60:
+            if time_since_last_input < 90:
+                # streamer should stop automatically after 60s, so give ourselves a 30s grace period to shutdown
+                logging.info(
+                    f"Detected DEGRADED_INPUT during shutdown: time_since_last_input={time_since_last_input:.1f}s"
+                )
+                return PipelineState.DEGRADED_INPUT
+            return PipelineState.OFFLINE
+        elif time_since_last_input > 2 or input.fps < 15:
```
emranemran, line 133: nit: should we make input.fps here configurable using an argument instead? We might end up shipping a dreamshaper workflow that runs right around 15-ish fps.

runner/app/pipelines/live_video_to_video.py
```diff
@@ -22,38 +22,63 @@ def __init__(self, model_id: str):
...
+        if not self.process:
+            raise RuntimeError("Pipeline process not running")
+
+        max_retries = 10
+        thrown_ex = None
+        for attempt in range(max_retries):
+            if attempt > 0:
+                time.sleep(1)
+            try:
+                conn = http.client.HTTPConnection("localhost", 8888)
```
varshith15, line 48: nit: using it as contextmanager (wraping it with `with`) will look cleaner
also probably don't need to use httpconnection, its a one time call -- so something light weight like requests should do
if it's be reused, maintain it as a class variable -- don't need to make a connection each time

varshith15: can pick it up later

emranemran (approved):

PR #427, opened 2025-02-16
latency logs

varshith15:
adds latency logs

### DIFF
--- a/runner/app/live/infer.py
+++ b/runner/app/live/infer.py
@@ -16,7 +16,7 @@ infer_root = os.path.abspath(os.path.dirname(__file__))
 sys.path.insert(0, infer_root)
 
 from api import start_http_server
-from log import config_logging
+from log import config_logging, log_timing
 from streamer.protocol.trickle import TrickleProtocol
 from streamer.protocol.zeromq import ZeroMQProtocol
 
@@ -53,19 +53,20 @@ async def main(
 
     api = None
     try:
-        await process.start()
-        if streamer:
-            await streamer.start(params)
-        api = await start_http_server(http_port, process, streamer)
+        with log_timing("[infer.py]: ProcessGuardian started successfully in"):
+            await process.start()
+            if streamer:
+                await streamer.start(params)
+            api = await start_http_server(http_port, process, streamer)
 
-        tasks: List[asyncio.Task] = []
-        if streamer:
-            tasks.append(asyncio.create_task(streamer.wait()))
-        tasks.append(
-            asyncio.create_task(block_until_signal([signal.SIGINT, signal.SIGTERM]))
-        )
+            tasks: List[asyncio.Task] = []
+            if streamer:
+                tasks.append(asyncio.create_task(streamer.wait()))
+            tasks.append(
+                asyncio.create_task(block_until_signal([signal.SIGINT, signal.SIGTERM]))
+            )
 
-        await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+            await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
     except Exception as e:
         logging.error(f"Error starting socket handler or HTTP server: {e}")
         logging.error(f"Stack trace:\n{traceback.format_exc()}")
--- a/runner/app/live/infer.py
+++ b/runner/app/live/infer.py
@@ -59,14 +59,14 @@ async def main(
                 await streamer.start(params)
             api = await start_http_server(http_port, process, streamer)
 
-            tasks: List[asyncio.Task] = []
-            if streamer:
-                tasks.append(asyncio.create_task(streamer.wait()))
-            tasks.append(
-                asyncio.create_task(block_until_signal([signal.SIGINT, signal.SIGTERM]))
-            )
+        tasks: List[asyncio.Task] = []
+        if streamer:
+            tasks.append(asyncio.create_task(streamer.wait()))
+        tasks.append(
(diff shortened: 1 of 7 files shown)


runner/app/live/infer.py
```diff
@@ -53,10 +53,11 @@ async def main(
 
     api = None
     try:
-        await process.start()
-        if streamer:
-            await streamer.start(params)
-        api = await start_http_server(http_port, process, streamer)
+        with log_timing("[infer.py]: ProcessGuardian started successfully in"):
```
emranemran, line 56: Are there other places where we could log like this? e.g. when request is first received?

emranemran:
How did testing look like? Any impact to resource usage (e.g. CPU usage or heavy disk i/o from logging) before/after this change? 

Also, can you post some examples of what the logs looks like?

runner/app/live/trickle/encoder.py
```diff
@@ -129,3 +134,17 @@ def rescale_ts(pts: int, orig_tb: Fraction, dest_tb: Fraction):
...
     return int(round(float((Fraction(pts) * orig_tb) / dest_tb)))
+
+
+def log_frame_timestamps(frame_type: str, frame: InputFrame):
+    ts = frame.log_timestamps
+    
+    def log_duration(start_key: str, end_key: str):
+        if start_key in ts and end_key in ts:
+            duration = ts[end_key] - ts[start_key]
+            logging.debug(f"{frame_type} {start_key} to {end_key} took {duration}s")
```
leszko, line 145: Do we print the debug logs in staging/prod? I tried it on my TensorDock dev and I haven't seen those logs.

runner/app/live/streamer/protocol/trickle.py
```diff
@@ -83,10 +83,7 @@ def dequeue_frame():
 
     async def egress_loop(self, output_frames: AsyncGenerator[OutputFrame, None]):
         def enqueue_bytes(frame: OutputFrame):
-            if frame:
-                self.publish_queue.put(frame)
-            else:
-                self.publish_queue.put(None)
+            self.publish_queue.put(frame)
```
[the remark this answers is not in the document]
leszko: Yeah, I'd also be interested. Maybe it's fine, just wonder, because someone put that IF 🙃

leszko (approved):
Added 2 comments, other than that, it looks good from my end.

I haven't tested the CPU impact, but I think it should be ok, we make a few logs per frame and we have 24 FPS, so this should not have huge impact IMO.

runner/app/live/trickle/encoder.py
varshith15: i remember manually setting VERBOSE_LOGGING as 1 to print debug logs -- not sure if there a way to deploy on staging with debug on

runner/app/live/streamer/protocol/trickle.py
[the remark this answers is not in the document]
varshith15: the egress loop will always return a non None value https://github.com/livepeer/ai-runner/blob/8c6db471375e5df3349f92a506824e80b74dc18e/runner/app/live/streamer/streamer.py#L214

runner/app/live/infer.py
varshith15: @victorges can you please have a look and see if any other logs are needed to give us a better picture of the startup time

varshith15:
@emranemran  
Screenshot 2025-02-18 at 8 42 23 PM

runner/app/live/streamer/process.py
```diff
@@ -119,13 +119,13 @@ def report_error(error_msg: str):
...
-                pipeline = load_pipeline(self.pipeline_name, **params)
-                logging.info("Pipeline loaded successfully")
+                with log_timing("PipelineProcess: Pipeline loaded successfully in"):
+                    pipeline = load_pipeline(self.pipeline_name, **params)
             except Exception as e:
                 report_error(f"Error loading pipeline: {e}")
                 try:
-                    pipeline = load_pipeline(self.pipeline_name)
-                    logging.info("Pipeline loaded with default params")
+                    with log_timing("PipelineProcess: Pipeline loaded with default params in"):
```
ecmulli, line 127:
```suggestion
                    with log_timing("PipelineProcess: Pipeline load successful with default params"):
```

runner/app/live/trickle/encoder.py
ecmulli, line 145:
```suggestion
            logging.debug(f"frame_type={frame_type} start_key={start_key} end_key={end_key} duration_s={duration}")
```

ecmulli (commented):
Overall, I love the functionality, but we need to ensure our logging is fully `logfmt` compliant.  

The key reason is that it's easy to convert a structured, machine-readable log into a human-readable format, but the reverse is much more difficult.  

For example, a `logfmt`-formatted log like: 
```
“frame_type={frame_type} start_key={start_key} end_key={end_key} duration_s={duration}”
```
can be effortlessly transformed in Loki:  
```
{ } | logfmt | {{.frame_type}} {{.start_key}} to {{.end_key}} took {{.duration_s}} seconds
```
However, a freeform log like:  
```
logging.debug(f”{frame_type} {start_key} to {end_key} took {duration}s”)
```
is much harder—if not impossible—to parse or transform reliably.  

Let me know what you think!

runner/app/live/infer.py
victorges: Oh btw we shouldn't add this `[infer.py]` prefix, it will get duplicated. That's added automatically on the code that starts the infer.py process.

PR #430, opened 2025-02-19
Bump comfy base image version

pschroedl:
Dockerfile.live-base-comfyui base image -> livepeer/comfyui-base:0.0.2-ai-runner

### DIFF
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -6,6 +6,7 @@ import queue
 import sys
 import time
 from typing import Any
+import torch
 
 from pipelines import load_pipeline
 from log import config_logging, config_logging_fields, log_timing
@@ -89,17 +90,25 @@ class PipelineProcess:
                 continue
         return None
 
-    def get_recent_logs(self, n=10) -> list[str]:
+    def get_recent_logs(self, n=None) -> list[str]:
+        """Get recent logs from the subprocess. If n is None, get all available logs."""
         logs = []
         while not self.log_queue.empty():
             try:
                 logs.append(self.log_queue.get_nowait())
             except queue.Empty:
                 break
-        return logs[-n:]  # Return last n logs
+        return logs[-n:] if n is not None else logs  # Only limit if n is specified
 
     def process_loop(self):
         self._setup_logging()
+
+        # Ensure CUDA environment is available inside the subprocess.
+        # Multiprocessing (spawn mode) does not inherit environment variables by default,
+        # causing `torch.cuda.current_device()` checks in ComfyUI's model_management.py to fail.
+        # Explicitly setting `CUDA_VISIBLE_DEVICES` ensures the spawned process recognizes the GPU.
+        if torch.cuda.is_available():
+            os.environ["CUDA_VISIBLE_DEVICES"] = str(torch.cuda.current_device())
 
         def report_error(error_msg: str):
             error_event = {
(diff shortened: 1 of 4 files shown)


leszko (changes requested):
My understanding after chatting with @varshith15 is that before merging this PR, we need to make some code changes. 

[Discord thread](https://discord.com/channels/423160867534929930/1332431145240887336/1342125869308837978)

runner/docker/Dockerfile.live-app__PIPELINE__
```diff
@@ -29,12 +29,8 @@ COPY bench.py /app/bench.py
 
 WORKDIR /app
 
-# Dirty hack to remove excessive logging before this PR is merged
-# https://github.com/hiddenswitch/ComfyUI/pull/28
-RUN sed -i 's/logging.info(f"Loaded {to_load}")//' /miniconda3/envs/comfystream/lib/python3.11/site-packages/comfy/model_management.py
-
 # Remove ComfyUI Manager node, because it is not needed for this application and it takes 10s additional time to start up
-RUN rm -r /ComfyUI/custom_nodes/ComfyUI-Manager
+RUN test -d /workspace/ComfyUI/custom_nodes/ComfyUI-Manager && rm -r /workspace/ComfyUI/custom_nodes/ComfyUI-Manager || true
```
victorges, line 33: An alternative is using `-f` on `rm` to not fail if the folder doesnt exist
```suggestion
RUN rm -rf /workspace/ComfyUI/custom_nodes/ComfyUI-Manager
```

victorges (approved):
LGTM!

victorges:
Were you able to test it locally E2E?

runner/app/live/streamer/process.py
```diff
@@ -89,18 +90,26 @@ async def recv_output(self) -> OutputFrame | None:
...
 
     def process_loop(self):
         self._setup_logging()
 
+        # Ensure CUDA environment is available inside the subprocess.
+        # Multiprocessing (spawn mode) does not inherit environment variables by default,
+        # causing `torch.cuda.current_device()` checks in ComfyUI's model_management.py to fail.
+        # Explicitly setting `CUDA_VISIBLE_DEVICES` ensures the spawned process recognizes the GPU.
+        if torch.cuda.is_available():
+            os.environ["CUDA_VISIBLE_DEVICES"] = str(torch.cuda.current_device())
```
emranemran, line 111: nice find! so if i understand correctly, in prior images, we weren't launching comfystream/comfyui using the multiprocessing module?

emranemran (approved):

runner/docker/Dockerfile.live-app__PIPELINE__
pschroedl: Removed

runner/app/live/streamer/process.py
pschroedl: We were using python multiprocessing, even prior to the PR to [Run a single pipeline process for all streams](https://github.com/livepeer/ai-runner/pull/424/files#top).  

I initially thought that may have been the culprit, but given that it is not a new change, it is possible that the check on the comfy side in model_management.py was changed in the hiddenswitch upgrade to rely on this environment variable.

eliteprox:
> > My understanding after chatting with @varshith15 is that before merging this PR, we need to make some code changes.
> > [Discord thread](https://discord.com/channels/423160867534929930/1332431145240887336/1342125869308837978)
> 
> I've updated the base container to one provided by @eliteprox that only contains the path changes, hiddenswitch upgrade, and face controlnet changes https://discord.com/channels/423160867534929930/1342638428822704149/1344056919735668818
> 
> This should not require additional code changes beyond what I've added to this PR.

Correct, this is a phased update from https://github.com/yondonfu/comfystream/tree/ai-runner containing only the changes above

PR #432, opened 2025-02-26
feat: comfyui stop

varshith15:
PR adds the functionality to stop comfyui when stopping the process

### DIFF
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -37,6 +37,7 @@ class PipelineProcess:
         self.start_time = 0.0
 
     async def stop(self):
+        #TODO: syncrounous call might be blocking, figure out a better way
         self._stop_sync()
 
     def _stop_sync(self):
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -24,7 +24,6 @@ class PipelineProcess:
 
     def __init__(self, pipeline_name: str):
         self.pipeline_name = pipeline_name
-        self.pipeline = None
         self.ctx = mp.get_context("spawn")
 
         self.input_queue = self.ctx.Queue(maxsize=5)
@@ -38,8 +37,6 @@ class PipelineProcess:
         self.start_time = 0.0
 
     async def stop(self):
-        if self.pipeline:
-            await self.pipeline.stop()
         await asyncio.to_thread(self._stop_sync)
 
     def _stop_sync(self):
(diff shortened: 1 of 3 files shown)


leszko (approved):

runner/app/live/streamer/process.py
```diff
@@ -37,6 +38,8 @@ def __init__(self, pipeline_name: str):
         self.start_time = 0.0
 
     async def stop(self):
+        if self.pipeline:
+            await self.pipeline.stop()
```
victorges, line 42: I think we'll need to move the stop call into the `processs_loop` otherwise this will not work. This function, even though is in the `Process` class, will run in the outer process to manage the pipeline process instead. We instead need to run the stop from inside the pipeline process itself. Putting in other words, this `self.pipeline` will likely be `None` in this context since it's set from within the process but here we're in the outer process.

runner/app/live/streamer/process.py
```diff
@@ -120,12 +123,12 @@ def report_error(error_msg: str):
 
             try:
                 with log_timing("PipelineProcess: Pipeline loaded successfully"):
-                    pipeline = load_pipeline(self.pipeline_name, **params)
+                    self.pipeline = load_pipeline(self.pipeline_name, **params)
```
victorges, line 126: Yes this will not work. Even if the shared var worked, we'd hve a concurrency issue here since one place is using the pipeline and the other would be stopping it.

varshith15: i missed the part about process_loop being a separate process, added the stop inside process_loop after the while loop is stopped

runner/app/live/streamer/process.py
varshith15: makes sense

runner/app/live/streamer/process.py
```diff
@@ -170,6 +171,15 @@ def report_error(error_msg: str):
                     report_error(f"Error processing frame: {e}")
         except Exception as e:
             report_error(f"Error in process run method: {e}")
+        finally:
+            self._cleanup_pipeline(pipeline)
```
victorges, line 175: This makes more sense to me! The only question I have is whether this finally block will actually run. I think it might be the case that processes get killed wherever they are in the stack and don't get to run finally blocks. It might not be the case, maybe we do get a stop exception or something. Nevertheless, make sure you test this to ensure this finally block runs when the process is restarted

victorges (approved):
LGTM but gotta make sure it works

runner/app/live/streamer/process.py
varshith15: how did you simulate a failure for your process gaurdian PR?

runner/app/live/streamer/process.py
```diff
@@ -37,10 +37,12 @@ def __init__(self, pipeline_name: str):
         self.start_time = 0.0
 
     async def stop(self):
-        await asyncio.to_thread(self._stop_sync)
+        self._stop_sync()
```
victorges, line 40: Can you add a TODO here to avoid calling synchronous stuff from an asynchronous context? It's ok for now but ideally we shouldn't be doing this

victorges (approved):
Neaet!

PR #405, opened 2025-01-29
docker: Use comfystream base image as base for comfyui pipeline

victorges:
This implements the [task](https://www.notion.so/livepeer/Migrate-ai-runner-docker-to-comfystream-base-image-1890a348568780c1ae9ef2ee26fced82?pvs=4) for reusing `comfystream` base docker image
to avoid setting up a separate environment altogether for the `comfyui` ai-runner.

PR #438, opened 2025-03-04
Fix dl_checkpoints script

mjh1:
Following on from https://github.com/livepeer/ai-runner/pull/437, the script still wasn't working for me locally or in CI so these changes should fix it.

### DIFF
--- a/runner/dl_checkpoints.sh
+++ b/runner/dl_checkpoints.sh
@@ -99,7 +99,7 @@ function download_live_models() {
   fi
   # ai-worker has tags hardcoded in `var livePipelineToImage` so we need to use the same tag in here:
   docker image tag $AI_RUNNER_COMFYUI_IMAGE livepeer/ai-runner:live-app-comfyui
-  docker run --rm -v ./models:/workspace/ComfyUI/models/ --gpus all -l ComfyUI-Setup-Models $AI_RUNNER_COMFYUI_IMAGE \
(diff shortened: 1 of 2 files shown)


runner/dl_checkpoints.sh
```diff
@@ -133,9 +133,9 @@ function build_tensorrt_models() {
 
   # Dreamshaper-8-Dmd-1kstep
   # TODO: Remove the script download with curl. It should already come in the base image once eliteprox/comfystream#1 is merged.
```
pwilczynskiclearcode, line 135: `TODO:` got confusing as now we don't use eliteprox's comfyui. Now it's yondon's.

There's https://github.com/eliteprox/comfystream/blob/main/src/comfystream/scripts/build_trt.py on the `eliteprox`. I guess it doesn't do what we want?

pwilczynskiclearcode: It looks it was confusing from the beginning https://github.com/livepeer/ai-runner/pull/405/files#diff-49045665a4e2315c6702f467b07ebeff183e1a7db386704f45a5486da05b1857R147

mjh1: Yeah I wasn't sure about my change to this version of `build_trt.py` @pschroedl could you check? I think we just needed a newer version with the `/workspace` prefix update.

PR #445, opened 2025-03-07
create cliptext model directory

pschroedl:
comfystream setup_models.py should ultimately handle these missing nested directories, but this should unbreak staging

### DIFF
--- a/runner/dl_checkpoints.sh
+++ b/runner/dl_checkpoints.sh
@@ -97,6 +97,8 @@ function download_live_models() {
     echo "ERROR: ComfyUI base image $AI_RUNNER_COMFYUI_IMAGE not found"
     return 1
   fi
+  # comfystream setup_models.py should ultimately handle these missing nested directories
+  mkdir -p ./models/text_encoders/CLIPText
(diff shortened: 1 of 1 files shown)


hjpotter92 (approved):

pschroedl:
I was just able to verify again that setup_models creates the text_encoders/CLIPText folders even when running from the docker container as happens in dl_checkpoints and deploy_models, so this probably isn't necessary ( or the issue behind this model not being found )

eliteprox:
> I was just able to verify again that setup_models creates the text_encoders/CLIPText folders even when running from the docker container as happens in dl_checkpoints and deploy_models, so this probably isn't necessary ( or the issue behind this model not being found )

Correct, the script creates the directories so no need to create missing ones

https://github.com/livepeer/comfystream/blob/a4c564d4e3eea7e7e5fbe801701c1f8b427ef6bc/src/comfystream/scripts/setup_models.py#L85-L94

https://github.com/livepeer/comfystream/blob/a4c564d4e3eea7e7e5fbe801701c1f8b427ef6bc/src/comfystream/scripts/setup_models.py#L25

PR #393, opened 2025-01-08
Remove liveportrait

leszko:

PR #461, opened 2025-03-19
deprecate: old pipelines

varshith15:
removes old live pipelines - streamdiffusion and sam2


victorges (approved):
Can you remove anything from the CI as well? Maybe use this PR as a reference for other stuff you might need to remove https://github.com/livepeer/ai-runner/pull/393/files#diff-dc7ed550a8dfb1d664c9b99b9fb8d6d04d11fc2d0f7d465bd524598de6b8cb59L86

PR #465, opened 2025-03-20
live pipeline Docker workflow to push images to repo on PRs

pwilczynskiclearcode:


.github/workflows/ai-runner-live-pipelines-docker.yaml
```diff
@@ -209,7 +208,7 @@ jobs:
         with:
           images: livepeer/ai-runner
           tags: |
-            type=raw,value=live-app-noop
+            type=raw,value=live-app-noop,enable={{is_default_branch}}
```
hjpotter92, line 211: are we using noop images anywhere? what's the use-case. otherwise we can save on some build hours on the runner by disabling these 

@eliteprox / @rickstaa

pwilczynskiclearcode: OK. Maybe let me remove `-noop` for now and we can reintroduce it if there's actually a case for deploying that into the secondary staging

varshith15: noop is just being used for local testing and nothing else

pwilczynskiclearcode: OK. so I'm just adding `,enable={{is_default_branch}}` (just in case) but don't enable docker pushes for `noop`

hjpotter92 (approved):

PR #475, opened 2025-03-25
Revert "Fix delayed teardowns"

varshith15:
Reverts livepeer/ai-runner#441

This didn't solve the issue. control_loop is not a compulsory loop.


j0sh:
@varshith15 There may be a couple different issues happening here because this _did_ solve an issue in my local environment.

Does it make anything worse?

PR #479, opened 2025-03-26
feat: comfystream v0.0.4 ai-runner port

varshith15:


runner/app/live/pipelines/loader.py
```diff
@@ -1,10 +1,10 @@
 from .interface import Pipeline
 
-def load_pipeline(name: str, **params) -> Pipeline:
+def load_pipeline(name: str) -> Pipeline:
     if name == "comfyui":
         from .comfyui import ComfyUI
-        return ComfyUI(**params)
```
eliteprox, line 6: `**params` are removed because they are passed into the pipeline after initialization from `process.py` using `pipeline.update_params` https://github.com/livepeer/ai-runner/pull/479/files?diff=unified&w=1#diff-884ec47d03a7fdce1ff7a0fc4f9858f45965ff1cbb3ee034c1f3505f756a7f2bR206.

runner/app/live/streamer/streamer.py
```diff
@@ -55,10 +55,10 @@ async def start(self, params: dict):
             run_in_background("ingress_loop", self.run_ingress_loop()),
             run_in_background("egress_loop", self.run_egress_loop()),
             run_in_background("report_status_loop", self.report_status_loop()),
-            run_in_background("control_loop", self.run_control_loop()),
         ]
         # auxiliary tasks that are not critical to the supervisor, but which we want to run
         self.auxiliary_tasks = [
+            run_in_background("control_loop", self.run_control_loop()),
```
eliteprox, line 61: This change makes the control_url optional which is important for manual testing, do we need to move it back to `self.main_tasks`? It would be convenient to keep this change https://github.com/livepeer/ai-runner/blob/a1e1b33654870e06377b71862bc15bddf6e63b99/runner/app/live/streamer/streamer.py#L54-L59

runner/app/live/streamer/process.py
```diff
@@ -117,91 +119,108 @@ def process_loop(self):
...
+    def _handle_logging_params(self, params: dict) -> dict:
+        if isinstance(params, dict) and "request_id" in params and "stream_id" in params:
+            logging.info(f"PipelineProcess: Resetting logging fields with request_id={params['request_id']}, stream_id={params['stream_id']}")
+            self._reset_logging_fields(
+                params["request_id"], params["stream_id"]
+            )
+            return {}
+        return params
+
+    async def _initialize_pipeline(self):
```
victorges, line 137: neat I like the breakdown in helper functions

runner/app/live/streamer/process.py
```diff
@@ -117,91 +119,108 @@ def process_loop(self):
...
-                        self.output_queue.put(output_frame)
-                    elif isinstance(input_frame, AudioFrame):
-                        self.output_queue.put(AudioOutput([input_frame]))
-                        # TODO wire in a proper pipeline here
-                    else:
-                        report_error(f"Unsupported input frame type {type(input_frame)}")
-                except Exception as e:
-                    report_error(f"Error processing frame: {e}")
+        try:
+            await asyncio.gather(input_task, output_task, param_task)
```
victorges, line 171: We should stop the whole thing if any of these tasks, otherwise we will continue running even with only part of the pipeline working, which is invalid. You can use `asyncio.wait` instead [like the streamer](https://github.com/livepeer/ai-worker/blob/29408e443488107eb9a33af06f01e6b0bbe46564/runner/app/live/streamer/streamer.py#L75)

Then on exceptions we should make sure we cancel the other tasks (or maybe just let them be, we're gonna exit the process anyway)

runner/app/live/streamer/process.py
```diff
@@ -117,91 +119,108 @@ def process_loop(self):
...
-        finally:
-            self._cleanup_pipeline(pipeline)
+            self._report_error(f"Error in pipeline loops: {e}")
+            await self._cleanup_pipeline(pipeline)
+            raise
+
+    async def _input_loop(self, pipeline):
+        while not self.is_done():
+            try:
+                input_frame = await asyncio.to_thread(self.input_queue.get, timeout=0.1)
```
victorges, line 180: maybe we dont need the timeout if we're not doing anything on `queue.Empty`? i think i only did it with a timeout myself when we were in a non-async context and i didn't want to block the thread (or maybe it was even before i knew `asyncio.to_thread` existed)

runner/app/live/streamer/process.py
```diff
@@ -117,91 +119,108 @@ def process_loop(self):
...
+                output_frame = await pipeline.get_processed_video_frame()
+                output_frame.log_timestamps["post_process_frame"] = time.time()
+                await asyncio.to_thread(self.output_queue.put, output_frame)
+            except Exception as e:
+                self._report_error(f"Error processing output frame: {e}")
+
+    async def _param_update_loop(self, pipeline):
+        while not self.is_done():
+            try:
+                params = self.param_update_queue.get_nowait()
```
victorges, line 203: We could probably have an `asyncio.to_thread` with no timeout here as well

runner/app/live/streamer/process.py
victorges: also side comment, i hope `asyncio` is really efficient and we'll not be creating a bunch of threads here

runner/app/live/streamer/streamer.py
victorges: Yeah I even though I couldn't really understand the need for the change that made it an aux task [1], let's keep the `control_url` on `main_tasks` to avoid unexpected regressions. 

[1] Josh mentioned it made the shutdown more efficient, which i think could be because the control channel is closed before the others on stream close

runner/docker/Dockerfile.live-app__PIPELINE__
```diff
@@ -11,13 +13,16 @@ RUN apt-get update && apt-get install -y \
...
 
 # Install any additional Python packages
-
 COPY requirements.live-ai.txt /app/requirements.txt
-RUN pip install --no-cache-dir -r /app/requirements.txt
-# TODO: Figure out a way to have this in requirements file
-RUN pip install --no-cache-dir triton==3.1.0
-RUN pip uninstall -y onnx onnxruntime onnxruntime-gpu
-RUN pip install onnx==1.17.0 onnxruntime-gpu==1.17.0
+COPY requirements.live.comfyui.txt /app/requirements.comfyui.txt
```
victorges, line 17: I think these could be directly in the `requirements.live-ai` for simplicity. We don't have any other uses of the `live-ai` deps anymore if not the comfyui image itself. Even the noop pipeline has its own `live-app-noop` dockerfile with hardcoded dependencies installed.

Is there any other reason for these new deps to be separate?

runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=livepeer/comfyui-base:ai-runner-9166aa9
+ARG BASE_IMAGE=livepeer/comfyui-base:latest
```
victorges, line 1: Can we maintain a fix image version here? We don't want this to update without us realizing, even if we know we can stay on the bleeding edge of `comfystream`. We'll always want to manually update this, test and only then deploy consciously to production. In the future we'd be able to automate most of this process as well. Ref: [[ai-runner devx phase 3]](https://discord.com/channels/423160867534929930/1332431145240887336/1342205704320323716)

runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -81,6 +81,5 @@ RUN rm -rf /workspace/ComfyUI/output && ln -s /models/ComfyUI--output /workspace
 RUN echo "source /workspace/miniconda3/etc/profile.d/conda.sh && conda activate comfystream" >> /conda_activate.sh && \
     chmod +x /conda_activate.sh
 ENV BASH_ENV=/conda_activate.sh
-SHELL ["/bin/bash", "-c"]
```
victorges, line 84: neat, love not needing this lol

runner/requirements.live-ai.txt
```diff
@@ -6,7 +6,7 @@ Pillow==10.3.0
 python-multipart==0.0.9
 uvicorn==0.34.0
 huggingface_hub==0.23.2
-triton>=2.1.0
+triton>=3.1.0
```
victorges, line 9: Oof, I remember having a hard time because fo this `triton` dep version. Make sure nothing breaks (I recall maybe it was catastrophic like it didn't even startup. so if you can get a stream going it should be fine)

runner/run-lv2v.sh
victorges, line 1: Has this script still've been useful for testing? Otherwise we could as well remove it.

victorges (approved):
LGTM! I tried to be very thorough since it is a big change, but most of the comments are just ideas or nits. Some long wanted refactors included here 👏

runner/app/live/pipelines/comfyui.py
```diff
@@ -263,46 +265,54 @@ def validate_prompt(cls, v) -> dict:
...
+        dummy_frame = VideoFrame(None, 0, 0)
+        dummy_frame.side_data.input = torch.randn(1, 512, 512, 3)
 
-        # Comfy will cache nodes that only need to be run once (i.e. a node that loads model weights)
-        # We can run the prompt once before actual inputs come in to "warmup"
-        warmup_input = torch.randn(1, 512, 512, 3)
-        asyncio.get_event_loop().run_until_complete(self.client.queue_prompt(warmup_input))
+        for _ in range(WARMUP_RUNS):
+          self.client.put_video_input(dummy_frame)
+          _ = await self.client.get_video_output()
```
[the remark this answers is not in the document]
varshith15: currently the code is written for 1:1 but it will take very minimal changes (have to figure out how to assign pts to the processed frame) for other

runner/app/live/streamer/process.py
```diff
@@ -74,6 +74,8 @@ def reset_stream(self, request_id: str, stream_id: str):
         # We internally use the param update queue to reset the logging configs
         self.param_update_queue.put({"request_id": request_id, "stream_id": stream_id})
 
+    # TODO: Once audio is implemented, combined send_input with input_loop
+    # We don't need additional queueing as comfystream already maintains a queue
```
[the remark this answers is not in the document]
varshith15: this is just duplicate code honestly, its like maintaining another ingress loop here. once comfystream is stable we should maintain a single queue and maybe get rid of process.py completely and maintain only streamer.py -- part of a different discussion

runner/app/live/pipelines/comfyui.py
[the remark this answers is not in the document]
eliteprox: Comfystream still supports only 1:1 input/output per workflow. The methods now accept multiple concurrent workflows, though concurrent workflows are not fully supported due to the need for separate processes per workflow.

Relevant methods:
- `set_prompts`:  https://github.com/livepeer/comfystream/blob/9ff4b39febde3970f9ed00765f955fea73d5153b/src/comfystream/client.py#L24-L27
- `update_prompts`: https://github.com/livepeer/comfystream/blob/9ff4b39febde3970f9ed00765f955fea73d5153b/src/comfystream/client.py#L29-L35
- `run_prompt`: https://github.com/livepeer/comfystream/blob/9ff4b39febde3970f9ed00765f955fea73d5153b/src/comfystream/client.py#L37C15-L44 (calls queue prompt async)

runner/app/live/pipelines/comfyui.py
```diff
@@ -263,46 +265,54 @@ def validate_prompt(cls, v) -> dict:
...
+        self.client.put_video_input(frame)
+        await self.video_incoming_frames.put(frame)
 
-        # Process using ComfyUI pipeline
-        result_tensor = asyncio.get_event_loop().run_until_complete(self.client.queue_prompt(image_tensor))
+    async def get_processed_video_frame(self):
+        result_tensor = await self.client.get_video_output()
+        frame = await self.video_incoming_frames.get()
+        while frame.side_data.skipped:
+            frame = await self.video_incoming_frames.get()
```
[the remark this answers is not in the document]
eliteprox: `get_processed_video_frame` 

- The boolean `frame.side_data.skipped` indicates if the output video frame is video only or audio + video. `frame.side_data.skipped==True` indicates the output is without audio and is statically set to True in the ai-runner implementation due to the existing audio passthrough upstream in the live video pipeline.

runner/run-lv2v.sh
eliteprox: It's been useful to me as reference documentation, I don't use the script for testing. I think we could move the gernal info into this guide and expand with more detail https://github.com/livepeer/ai-runner/blob/bc91b4ebe1888758c4d87f60d4fcf54abbf331b4/docs/live-ai-local-dev.md. @varshith15 WDYT? Maybe we should create an issue to address this later.

runner/docker/Dockerfile.live-base-comfyui
eliteprox: Yes, but there is still an opportunity to remove https://github.com/livepeer/ai-runner/pull/479/files?w=0#diff-443cdad78f488f139f012984eac0c36f930e5ceb4ce1dd436f19c865e7ddc639R81-R83. The process that is starting ComfyUI is infer.py. This can likely be solved with a change to start infer.py in the comfystream environment `cmd = ["conda", "run", "-n", "comfystream", "python", str(self.infer_script_path)]` https://github.com/livepeer/ai-runner/blob/3de85765f05314b6f86a8633bc2a548ceb94ff7f/runner/app/pipelines/live_video_to_video.py#L109-L111

runner/docker/Dockerfile.live-base-comfyui
eliteprox: Updated to fixed version matching v0.0.5 tag which this PR has been tested with https://github.com/livepeer/comfystream/releases/tag/v0.0.5

runner/app/live/streamer/streamer.py
eliteprox: Moved back to main_tasks https://github.com/livepeer/ai-runner/commit/aa845528a4b04236de7705d4f916a88c0d61224a. 
Maybe we can have a "switch" to turn this on/off for testing later or mock the control_url using a testing script without requiring go-livepeer to generate a control stream

runner/docker/Dockerfile.live-app__PIPELINE__
eliteprox: Initially I created a separate requirements.txt file because there are so many pinned version dependencies in the current one that it was breaking the comfyui node installation. Some packages like fastapi are still needed for the runner's own API server.

The strictly pinned versions are risky imo and should be handled in [setup_nodes.py](https://github.com/livepeer/comfystream/blob/main/src/comfystream/scripts/setup_nodes.py) and [nodes.yaml](https://github.com/livepeer/comfystream/blob/main/configs/nodes.yaml). In general, I think we should avoid pinning or changing certain kinds of python package versions in ai-runner and lean as close as possible to a clean output from setup_nodes.py. 

I'll test some changes here to clean this up. It's also related to https://github.com/livepeer/ai-runner/pull/479#discussion_r2032047865

runner/app/live/pipelines/comfyui.py
```diff
@@ -263,46 +265,54 @@ def validate_prompt(cls, v) -> dict:
...
-        super().__init__(**params)
-
+    def __init__(self):
         comfy_ui_workspace = os.getenv(COMFY_UI_WORKSPACE_ENV)
         self.client = ComfyStreamClient(cwd=comfy_ui_workspace)
         self.params: ComfyUIParams
+        self.video_incoming_frames = asyncio.Queue()
 
-        self.update_params(**params)
+    async def warm_video(self):
```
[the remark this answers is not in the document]
eliteprox: Added here https://github.com/livepeer/ai-runner/pull/479/commits/86f4f9723eb1416a79ff5a8227f934a4cb814ceb

runner/docker/Dockerfile.live-base-comfyui
eliteprox: I was able to solve this by starting infer.py with the comfystream conda env https://github.com/livepeer/ai-runner/pull/479/commits/2e046bb50ab54d69806f1958c19c6d1ac445e759

runner/app/live/streamer/process.py
eliteprox: Updated with asyncio.wait https://github.com/livepeer/ai-runner/blob/145f5ca7a9a2ab63c549764a42d4bdc4e815a8b3/runner/app/live/streamer/process.py#L170-L191

runner/app/live/streamer/process.py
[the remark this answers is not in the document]
victorges: Got it. That idea makes sense though for the future of comfystream 👍

runner/app/live/streamer/process.py
eliteprox: Addressed in https://github.com/livepeer/ai-runner/commit/d6922ab24ea8a5e22e3d565500404273be361ad2

runner/app/live/streamer/process.py
eliteprox: Addressed in https://github.com/livepeer/ai-runner/commit/d6922ab24ea8a5e22e3d565500404273be361ad2

runner/requirements.live-ai.txt
eliteprox: I noticed there was a difference in versions. I've reverted it back to triton>=2.1.0, but livepeer/comfyui-base will come with `Version: 3.2.0`. Ideally would like to minimize the python package installs here to minimize risk of breaking a custom node

runner/app/live/pipelines/interface.py
```diff
@@ -22,21 +23,37 @@ def __init__(self, **params):
...
+        """Get a processed frame from the pipeline.
 
         Returns:
-            Processed PIL Image
+            Processed VideoFrame
+        """
+        pass
+
+    @abstractmethod
+    async def set_params(self, **params):
```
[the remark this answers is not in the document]
eliteprox: Renamed to `initialize` in the interface https://github.com/livepeer/ai-runner/pull/479/commits/566c908ac5fc350145febe9ffdb9a2c8347b4d96

eliteprox:
@victorges I've addressed all review comments, will redeploy to secondary staging now to validate. Let me know if you see any issues

.github/workflows/ai-runner-live-pipelines-docker.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
         uses: tj-actions/changed-files@823fcebdb31bb35fdf2229d9f769b400309430d0 # v46.0.3
         with:
           files: |
-            runner/docker/Dockerfile.live-base
+            runner/docker/Dockerfile.live-base*
```
eliteprox, line 42: @pschroedl I don't think we need this change
```suggestion
            runner/docker/Dockerfile.live-base
```

runner/app/live/pipelines/interface.py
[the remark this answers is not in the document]
victorges: Neat!

runner/requirements.live-ai.txt
victorges: Does it work on triton 3.2.0? If it dos, feel free to remove it from here. This file is only installed on the comfyui image.

runner/app/live/streamer/process.py
```diff
@@ -122,93 +124,118 @@ def process_loop(self):
...
+                    await asyncio.to_thread(self.output_queue.put, AudioOutput([input_frame], self.request_id))
+            except queue.Empty:
+                continue
+            except Exception as e:
+                self._report_error(f"Error processing input frame: {e}")
+
+    async def _output_loop(self, pipeline):
+        while not self.is_done():
+            try:
+                output_frame = await pipeline.get_processed_video_frame(self.request_id)
```
victorges, line 209: Would there be a way to tag the images with a request_id on the input side instead? Otherwise I worry that we'll end up having the same issue we had before with the queues from the ai-runner code: frames from previous streams leaking to the next stream because of these queues, causing encoder/decoder crashes.

So if there could be a way we could "tag" the frame on the [input side on `put_video_frame`](https://github.com/livepeer/ai-runner/pull/479/files#diff-884ec47d03a7fdce1ff7a0fc4f9858f45965ff1cbb3ee034c1f3505f756a7f2bR198) we could actually track where the frame is coming from. Do you think that could work easily? Maybe on that `side_data` field. If it's not a simple change don't worry, this is more like a nice-to-have.

victorges (approved):
LGTM!

runner/app/live/streamer/process.py
eliteprox: I was concerned about that too, also considered using `side_data` to pass around the `request_id`, I read that it needed to remain consistent with the av frame type, wasn't fully sure if this comment is referring to VideoFrame or something else. Maybe we can create an issue for the backlog and @varshith15 can shed some light on this constraint?

victorges: What if instead of putting `VideoFrame`s on the `video_incoming_frames` we put `VideoOutput` objects instead? Then we could have the request_id in them passed through the queue here

runner/docker/Dockerfile.live-base-comfyui
```diff
@@ -8,6 +8,7 @@ FROM ${BASE_IMAGE}
 ENV DEBIAN_FRONTEND=noninteractive
 ENV NVIDIA_VISIBLE_DEVICES=all
 ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility,video
+ENV PATH="/workspace/miniconda3/bin:${PATH}"
```
pwilczynskiclearcode, line 11: Some of the changes in this PR caused `dl_checkpoints.sh` to [fail](https://github.com/livepeer/ai-runner/actions/runs/14355264860/job/40243134959):

```
ModuleNotFoundError: No module named 'yaml'
...
ModuleNotFoundError: No module named 'psutil'
```

eliteprox: This change was made early to cleanup environment activation, but it should be reverted back

eliteprox: Another reason I removed it is there isn't a python env at `/workspace/ComfyUI` so this line didn't seem to make any sense and shouldn't have directly cause the issue. More likely it was the the change below on ln 80. https://github.com/livepeer/ai-runner/pull/479/files/a90f5b79cc5738021db37a4251c68e1029e170b1#diff-443cdad78f488f139f012984eac0c36f930e5ceb4ce1dd436f19c865e7ddc639L80. This change was made because it caused issues and seemed like a "hacky" way to activate the env, occasionally it would activate the env twice when working in a dev container and entirely prevented deactivation.

Model compilation should happen inside of the comfystream conda env so it's using the same packages as the runtime env

PR #482, opened 2025-03-26
lv2v/infer.py: Set uncaught exception handlers to terminate process

victorges:
This creates custom uncaught exception handlers for both asyncio and threading modules,
to make sure we crash the whole process when our code fails to handle an exception.

This is important to catch problems where we are forgetting to handle exceptions and part
of the runner stop working without notice. On the worst case when those exceptions are not
to be fatal, all we need to do is to add an try/except clause and log them. They should not be
uncaught. This is the default behavior in other envs like node and go and idk why python likes
to be the weird one.

### DIFF
--- a/runner/app/live/infer.py
+++ b/runner/app/live/infer.py
@@ -6,8 +6,8 @@ import signal
 import sys
 import os
 import traceback
+import threading
 from typing import List
-import logging
 
 from streamer import PipelineStreamer, ProcessGuardian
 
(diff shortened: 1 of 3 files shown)


runner/app/live/infer.py
```diff
@@ -21,6 +21,34 @@
 from streamer.protocol.zeromq import ZeroMQProtocol
 
 
+def asyncio_exception_handler(loop, context):
+    """
+    Handles unhandled exceptions in asyncio tasks, logging the error and terminating the application.
+    """
+    exception = context.get('exception')
+    logging.error(f"Caught unhandled exception in asyncio task: {exception}")
```
pwilczynskiclearcode, line 29:
```python
exception = context.get('exception')
logging.error(f"Caught unhandled exception in asyncio task", exc_info=exception)
os.exit(1)
```

victorges: Oh nice, didn't know `exc_info` could get the exception directly

j0sh (approved):

PR #485, opened 2025-03-27
Improve the teardown time

leszko:
Don't always wait 60s+30s for the runner teardown.

### DIFF
--- a/runner/app/live/streamer/process_guardian.py
+++ b/runner/app/live/streamer/process_guardian.py
@@ -24,6 +24,7 @@ class ProcessGuardian:
 
         self.process = None
         self.monitor_task = None
+        self.stream_alive = False
         self.status = PipelineStatus(
             pipeline=pipeline, start_time=time.time()
         ).update_params(params, False)
@@ -47,6 +48,9 @@ class ProcessGuardian:
         if self.process:
             await self.process.stop()
             self.process = None
+    
+    def stop_stream(self):
+        self.stream_alive = False
 
     async def reset_stream(
         self,
@@ -55,6 +59,7 @@ class ProcessGuardian:
         params: dict,
         monitoring_callback: Callable[[dict], Awaitable[None]],
     ):
+        self.stream_alive = True
         if not self.process:
             raise RuntimeError("Process not running")
         self.status = PipelineStatus(pipeline=self.pipeline, start_time=time.time())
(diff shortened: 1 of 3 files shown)


mjh1 (approved):
Looks good but I'm definitely curious to see what @victorges thinks, maybe part of the existing logic is meant to be handling this but the timeouts are just set too high

runner/app/live/streamer/process_guardian.py
```diff
@@ -122,6 +127,9 @@ def _current_state(self) -> str:
         input = self.status.input_status
         last_input_time = input.last_input_time or 0
         time_since_last_input = current_time - last_input_time
+        # Stream is done and 10s grace period reached
+        if not self.stream_alive and time_since_last_input > 10:
```
victorges, line 131: The grace period was mostly to give time to the `_do_stop` to run (sometimes closing trickle stream can take a while for example). Since we are calling `stop_stream` in the end of `_do_stop`, we can either remove the grace period or make it much shorter, maybe 1-5s.

runner/app/live/streamer/process_guardian.py
```diff
@@ -47,6 +48,9 @@ async def stop(self):
         if self.process:
             await self.process.stop()
             self.process = None
+    
+    def stop_stream(self):
+        self.stream_alive = False
```
victorges, line 53: **supernit**: WDYT of naming this `stream_stopped` so it's more like a notification rather than an action? Otherwise it looks like this is actually going to stop something.

victorges (approved):
LGTM!

runner/app/live/streamer/process_guardian.py
leszko: ok, changed to 5s

runner/app/live/streamer/process_guardian.py
leszko: ok, updated

runner/app/live/streamer/process_guardian.py
```diff
@@ -45,6 +46,9 @@ async def stop(self):
         if self.process:
             await self.process.stop()
             self.process = None
+    
+    def stop_stream(self):
+        self.stream_stopped = False
```
victorges, line 51: @leszko is this boolean inverted? Maybe rename to `stream_running`

PR #487, opened 2025-03-28
live/process: Fail fast when pipeline can't be loaded

victorges:
When loading the process with default params we might end up with a `pipeline=None`
since the exception handler will only log the exception in case the initial params are already
empty (which means default).

This fixes by just re-raising the exception and crashing the process when the default params
fail to load.


varshith15 (approved):
LGTM! I missed the error raising

PR #493, opened 2025-03-31
Tensorrt RTX 4090 models in docker

pwilczynskiclearcode:
In here we just build a docker image with huggingface models:
* batch - we just pull the models and save in the image
* live - we pull the models and run tensorrt for a given GPU

In this PR I don't bake it into ai-runner:live... image yet. I've started to combine these two on a separate [PR](https://github.com/livepeer/ai-runner/pull/502/files)

### DIFF
--- a/.github/workflows/ai-runner-models.yaml
+++ b/.github/workflows/ai-runner-models.yaml
@@ -1,0 +1,132 @@
+name: Pull and dockerize huggingface models
+
+on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - "runner/dl_checkpoints.sh"
+      - ".github/workflows/ai-runner-models.yaml"
+  pull_request:
+    paths:
+      - "runner/dl_checkpoints.sh"
+      - ".github/workflows/ai-runner-models.yaml"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: ${{ !((github.ref_type == 'branch' && github.ref_name == 'main') || github.ref_type == 'tag') }}
+
+jobs:
+  download_batch_models:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          # Check https://github.com/livepeer/go-livepeer/pull/1891
+          # for ref value discussion
+          ref: ${{ github.event.pull_request.head.sha || '' }}
+
+      - name: dl_checkpoints.sh
+        env:
+          AI_RUNNER_COMFYUI_IMAGE: "livepeer/ai-runner:live-app-comfyui"
+        run: |
+          set -e
+          pip install 'huggingface_hub[cli,hf_transfer]' hf_transfer
+          runner/dl_checkpoints.sh --batch
+          du -hs ./models/
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.CI_DOCKERHUB_USERNAME }}
+          password: ${{ secrets.CI_DOCKERHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            livepeer/ai-runner-models
+          tags: |
+            type=sha,prefix=batch-
+            type=ref,event=pr,prefix=batch-
+            type=ref,event=tag,prefix=batch-
+            type=sha,format=long,prefix=batch-
+            type=ref,event=branch,prefix=batch-
+            type=raw,value=latest-batch,enable={{is_default_branch}}
+
+      - name: Build and push image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          platforms: linux/amd64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          file: runner/docker/Dockerfile.models
+          cache-from: type=registry,ref=livepeerci/ai-runner-models:dockerbuildcache-batch
+          cache-to: type=registry,ref=livepeerci/ai-runner-models:dockerbuildcache-batch,mode=max
+
+  download_build_tensorrt_models:
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/ai-runner-models.yaml
```diff
@@ -0,0 +1,132 @@
...
+      - "runner/dl_checkpoints.sh"
+      - ".github/workflows/ai-runner-models.yaml"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: ${{ !((github.ref_type == 'branch' && github.ref_name == 'main') || github.ref_type == 'tag') }}
+
+jobs:
+  download_batch_models:
```
pwilczynskiclearcode, line 21: batch just [downloads the models](https://github.com/livepeer/ai-runner/blob/main/runner/dl_checkpoints.sh#L159-L163) so it doesn't need to be built per GPU.

.github/workflows/ai-runner-models.yaml
```diff
@@ -0,0 +1,132 @@
...
+          tags: ${{ steps.meta.outputs.tags }}
+          file: runner/docker/Dockerfile.models
+          cache-from: type=registry,ref=livepeerci/ai-runner-models:dockerbuildcache-batch
+          cache-to: type=registry,ref=livepeerci/ai-runner-models:dockerbuildcache-batch,mode=max
+
+  download_build_tensorrt_models:
+    strategy:
+      fail-fast: false
+      matrix:
+        gpu: [rtx-4090]
```
pwilczynskiclearcode, line 79: Just being prepared for other architectures

runner/docker/Dockerfile.models
```diff
@@ -0,0 +1,3 @@
+FROM alpine:3
+
+COPY ./models/ /models/
```
pwilczynskiclearcode, line 3: Ideally I'd write sth like:

```
FROM livepeer/ai-worker:live-app-comfyui AS builder
RUN runner/dl_checkpoints.sh --tensorrt

FROM alpine:3
COPY --from:builder /models/ /models/
```

.github/workflows/ai-runner-models.yaml
```diff
@@ -0,0 +1,132 @@
...
+    branches:
+      - "main"
+    paths:
+      - "runner/dl_checkpoints.sh"
+      - ".github/workflows/ai-runner-models.yaml"
+  pull_request:
+    paths:
+      - "runner/dl_checkpoints.sh"
+      - ".github/workflows/ai-runner-models.yaml"
+  workflow_dispatch:
```
pwilczynskiclearcode, line 14: We just want to redownload/rebuild the models when `dl_checkpoints.sh` changed or on demand.

I think that this should be independent from building ai-runner containers which are based on our code changes.

victorges: The tricky part is that the `dl_checkpoints.sh` script uses the `ai-runner` image to run the tensorrt step, so there's kind of a cyclic dependency here. But we could have a caveat documented somewhere that one needs to `workflow_dispatch` this manually when the runner image changes in a significant way.

runner/docker/Dockerfile.models
victorges, line 1: Could you explain the reasoning behind the higher level change? Like why are we embedding the models in the docker images (especially in the runner image itself). I think I was not in the discussions where this decision was made.

victorges (approved):
LGTM

runner/docker/Dockerfile.models
hjpotter92: it could just be
```dockerfile
FROM scratch
COPY ...
```
no need to even go for alpine until we start building the models inside docker itself.

.github/workflows/ai-runner-models.yaml
hjpotter92: both jobs could be made into a single strategy matrix, so that any future common change doesnt need to be copied over.

pwilczynskiclearcode:
> 1. Do we have a plan if we have different GPUs? Will we build a separate Docker image for each GPU?

Yes. The [gpu matrix](https://github.com/livepeer/ai-runner/pull/493/files#diff-70cbc35611aba686d7d53a20cafe0146fcfbe124323b2c02154cc0925895f686R78-R79) is just a single `rtx-4090` architecture. For know I'm just making sure the naming conventions are prepared for different architectures. Yes. We'd need to build it for each architecture using separate runners.

> 2. Currently we have more problems with pulling docker images than with downloading/extracting tarballs, so I wonder what's the reason for this change?

We do have problems with tarballs from Bunny CDN. It also happened that they were downloaded really long so it might be a general Tensordock thing. After unpacking .tar we can't guarantee it's content consistency. With docker we can easier manage versioning, caching layers, etc. Also we can guarantee that both models and the code are updated atomically (all or nothing) as currently we first pull and unpack models and then independently restart the runner.

With docker image models are just one layer that if not changes should be cached as long as it doesn't change.

Containerizing models give us assurance that the same thing that was tested on staging gets delivered unchanged to production.

One last thing is that if we consider GPU providers like salad.com or other fully containerised - so this approach prepares us for this scenario as well.

> 3. One potential disadvantage is that the Docker build process will take even longer. Do we know how much longer will it build?

Here I'm introducing models building workflow as separate from regular ai-worker builds. It should be built only once in a while - when huggingface models update (rarely) or if we introduce changes to `dl_checkpoints.sh`, `setup_models.py`, `export_trt.py` or `build_trt.py`.
We'd have it later on just pasted into ai-runner image with: [`COPY --from=livepeer/ai-runner-models-tensorrt-rtx-4090 /models /models`](https://github.com/livepeer/ai-runner/pull/502/files#diff-96b0c92168123272ec344fb0d23f0c361e9f819b661fa61a416d759c816f9dafR6) which (if used with proper docker and github caching) shouldn't be too time consuming. There's another possibility. Instead of [baking models into ai-runner](https://github.com/livepeer/ai-runner/pull/502/files#diff-96b0c92168123272ec344fb0d23f0c361e9f819b661fa61a416d759c816f9dafR6) we could deliver and run them separately just mount the models cross-container.

runner/docker/Dockerfile.models
pwilczynskiclearcode: I [responded to](https://github.com/livepeer/ai-runner/pull/493#issuecomment-2775652163) Rafał questions and discussions started [on discord](https://discord.com/channels/423160867534929930/1346898307905093752/1346917063045812374)

leszko (approved):

PR #496, opened 2025-04-01
ai-runner-*pipelines-docker github workflows - update github runner ubuntu

pwilczynskiclearcode:
Screenshot 2025-04-01 at 17 21 35

### DIFF
--- a/.github/workflows/ai-runner-docker.yaml
+++ b/.github/workflows/ai-runner-docker.yaml
@@ -28,14 +28,6 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
-      - name: Check out code
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          # Check https://github.com/livepeer/go-livepeer/pull/1891
-          # for ref value discussion
-          ref: ${{ github.event.pull_request.head.sha }}
-
       - name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


.github/workflows/ai-runner-docker.yaml
```diff
@@ -28,14 +28,6 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - name: Check out code
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          # Check https://github.com/livepeer/go-livepeer/pull/1891
-          # for ref value discussion
-          ref: ${{ github.event.pull_request.head.sha }}
-
```
pwilczynskiclearcode, line 38: `docker/build-push-action@v6` is configured with `context: "{{defaultContext}}:runner"` meaning that it reads from git directly without looking at the local folder changes.

.github/workflows/ai-runner-live-pipelines-docker.yaml
```diff
@@ -165,7 +165,7 @@ jobs:
   build-noop:
     name: Build pipeline image (noop)
     needs: build-common-base
-    runs-on: ubuntu-20.04
```
hjpotter92, line 168: can instead use ubuntu-latest here.

PR #498, opened 2025-04-02
Docker image optimisation

pwilczynskiclearcode:
I believe that not all of these Dockerfiles/images are being used but let's introduce unified optimisations everywhere.

Proper optimisations would include separating heavy builds requiring additional system packages into separate Docker stages. But that's more complex task

### DIFF
--- a/runner/docker/Dockerfile.live-base
+++ b/runner/docker/Dockerfile.live-base
@@ -6,18 +6,18 @@ ENV NVIDIA_VISIBLE_DEVICES=all
 ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility,video
 
 # Install prerequisites
-RUN apt-get update && \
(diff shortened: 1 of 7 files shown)


runner/docker/Dockerfile.live-base
```diff
@@ -61,10 +60,7 @@ RUN cd ffmpeg && \
                 --disable-doc && \
     make -j$(nproc) && \
     make install && \
-    make distclean
-
-# Copy the compiled FFmpeg binaries to /usr/local
-RUN cp -R /compiled/* /usr/local/
-
-# Clean up APT cache to reduce image size
-RUN apt-get clean && rm -rf /var/lib/apt/lists/*
```
pwilczynskiclearcode, line 70: docker has incremental build so a cleanup operation like that must be under the same `RUN` where we install apt-get packages

hjpotter92 (approved):

PR #499, opened 2025-04-02
dl_checkpoints.sh - use built-in build_trt.py instead of curling it

pwilczynskiclearcode:


runner/dl_checkpoints.sh
```diff
@@ -132,10 +132,8 @@ function build_tensorrt_models() {
     )
 
   # Dreamshaper-8-Dmd-1kstep
-  # TODO: Remove the script download with curl. It should already come in the base image once eliteprox/comfystream#1 is merged.
```
pwilczynskiclearcode, line 135: https://github.com/eliteprox/comfystream/blob/main/src/comfystream/scripts/build_trt.py

hjpotter92 (approved):

PR #501, opened 2025-04-02
dl_checkpoints.sh - pull ai-runner image only if needed

pwilczynskiclearcode:


runner/dl_checkpoints.sh
```diff
@@ -92,6 +90,8 @@ function download_all_models() {
 
 # Download models only for the live-video-to-video pipeline.
 function download_live_models() {
+  docker pull "${AI_RUNNER_COMFYUI_IMAGE}"
```
pwilczynskiclearcode, line 93: Pull the runner image only if necessary (`--live` or `--tensorrt`).

pwilczynskiclearcode: I'm thinking of rewriting `dl_checkpoints.sh` so instead of running `docker run` the script would be run entirely inside docker

victorges (approved):

PR #503, opened 2025-04-02
pytest workflow - cache pip dependencies

pwilczynskiclearcode:


.github/workflows/ai-runner-test.yaml
```diff
@@ -31,4 +32,4 @@ jobs:
 
       - name: Run tests
         working-directory: runner
-        run: pytest
+        run: pytest --verbose --showlocals
```
hjpotter92, line 35: can also add step for code coverage through codecov.

PR #506, opened 2025-04-03
Reinitialize loops for each stream (fix AV Encoder Crash)

leszko:
Clear queues each time there is a new stream.

I _believe_ this PR will fix https://linear.app/livepeer/issue/ENG-2582/runner-encoder-av-crash (Runner Encoder AV Crash). This is what happens:
1. Sometimes the queues are not cleared
2. Then, during the next stream the first read frame is from the previous stream
3. Then, encoder crashes, because the timestamps do not match

### DIFF
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -281,4 +281,4 @@ def clear_queue(queue):
         try:
             queue.get_nowait()  # Remove items without blocking
         except Exception as e:
-            print(f"Error while clearing queue: {e}")
+            logging.error(f"Error while clearing queue: {e}")
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -71,7 +71,11 @@ class PipelineProcess:
         self.param_update_queue.put(params)
 
     def reset_stream(self, request_id: str, stream_id: str):
-        # We internally use the param update queue to reset the logging configs
+        clear_queue(self.input_queue)
+        clear_queue(self.output_queue)
+        clear_queue(self.param_update_queue)
+        clear_queue(self.error_queue)
+        clear_queue(self.log_queue)
         self.param_update_queue.put({"request_id": request_id, "stream_id": stream_id})
 
     def send_input(self, frame: InputFrame):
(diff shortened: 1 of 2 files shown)


runner/app/live/streamer/process.py
```diff
@@ -71,7 +71,11 @@ def update_params(self, params: dict):
         self.param_update_queue.put(params)
 
     def reset_stream(self, request_id: str, stream_id: str):
-        # We internally use the param update queue to reset the logging configs
+        clear_queue(self.input_queue)
+        clear_queue(self.output_queue)
+        clear_queue(self.param_update_queue)
+        clear_queue(self.error_queue)
```
varshith15, line 77: Maybe we should push the errors and logs instead of clearing the queue?

leszko: Hmm, I believe that if the given queue:
- is related to a stream (like `input_queue` or `output_queue`) then it's all good, we just clean it up for the safety measure
- is related to the infer process, then we should not clear it here at all

But I'm not sure if these `error_queue` or `logs_queue` is related to the process or a single stream? Wdyt @varshith15 @victorges

victorges: We actually can't easily push them at this point anymore since we don't have trickle streams. We'd need to change the logic to happen on the end of the stream instead of here (which is more like the beginning of the next)

So I think it could be tricky to do it, may e worth not having it for now

runner/app/live/streamer/protocol/trickle.py
```diff
@@ -64,8 +66,10 @@ async def stop(self):
         self.publish_task = None
 
     async def ingress_loop(self, done: asyncio.Event) -> AsyncGenerator[InputFrame, None]:
+        subscribe_queue = self.subscribe_queue
+        publish_queue = self.publish_queue
```
victorges, line 70: What's this change doing? I imagine it was from a previous version where you actually created new queues?

victorges (approved):
LGTM

runner/app/live/streamer/protocol/trickle.py
leszko: Just to be on the self side, I assign them to separate variables in case we change `self.subscribe_queue` somewhere. Since these `self.*` can be accessed from different threads AFAIU

leszko:
@varshith15 I'm merging this one (to test it on staging), but if you have any other comments, I'll address them in a separate PR.

PR #517, opened 2025-04-07
make sure we run only one dl_checkpoints.sh at a time

pwilczynskiclearcode:
`dl_checkpoints.sh --tensorrt` runs directly on the single rtx-4090 server so make sure they won't interfere with other simultaneous builds as they don't run inside docker

### DIFF
--- a/.github/workflows/ai-runner-models.yaml
+++ b/.github/workflows/ai-runner-models.yaml
@@ -80,6 +80,9 @@ jobs:
       matrix:
         gpu: [rtx-4090]
     runs-on: ${{ matrix.gpu }}
+    concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/ai-runner-models.yaml
```diff
@@ -80,6 +80,9 @@ jobs:
       matrix:
         gpu: [rtx-4090]
     runs-on: ${{ matrix.gpu }}
+    concurrency:
+      group: download-build-tensorrt-models
```
hjpotter92, line 84: should the group also reference `matrix.gpu` var; so that if later we have more gpus, each can have a separate build running? not sure how job specific concurrency groups work.

PR #525, opened 2025-04-09
live/pipeline: Tag frames with the right request_id

victorges:
The request Id from a frame should come from the input side, not output. Otherwise
we might still be handling output frames from the previous stream as if they were from
the right one.

On the way to this I ended up creating a separate file for the default workflow as well.
Keeps the comfyui pipeline implementation cleaner. 

Also added a bunch of missing types. Let's avoid writing untyped code pls.


runner/app/live/pipelines/comfyui.py
```diff
@@ -15,226 +16,10 @@
...
-      "clip_name": "CLIPText/model.fp16.safetensors"
-    },
-    "class_type": "CLIPLoader"
-  }
-}
-""")
+
+_default_workflow_path = pathlib.Path(__file__).parent.absolute() / "comfyui_default_workflow.json"
+with open(_default_workflow_path, 'r') as f:
+    DEFAULT_WORKFLOW_JSON = json.load(f)
```
eliteprox, line 22: This is so much better, nice improvement

eliteprox:
@victorges Apologies for overlooking this PR. I merged a change to fix the missing `request_id` parameters in https://github.com/livepeer/ai-runner/pull/526. Can you rebase?

victorges:
@eliteprox rebased

eliteprox (approved):
LGTM! Tested with `gateway_request_id=1234` and did not observe any mismatched frames

PR #526, opened 2025-04-09
Fix noop pipeline

j0sh:
This seems to have broken in https://github.com/livepeer/ai-runner/pull/479

Even with this PR, it still doesn't run due to this line expecting a hard-coded path to python that doesn't exist:

https://github.com/livepeer/ai-runner/blob/9e21e45480bf3112388f5b4b6bfb11df3ea51522/runner/app/pipelines/live_video_to_video.py#L110

When I change the executable back to `python` for local testing then it works fine, but presumably it was changed for a reason so I am not sure the best way to sort that out, maybe by updating the PATH env var in the container instead of hard-coding it here.


eliteprox (approved):
LGTM, thanks for catching this. I mistakenly assumed we had automated tests for the noop pipeline

eliteprox:
Added a potential fix here, feel free to cherry-pick if it looks good to you https://github.com/livepeer/ai-runner/pull/528

PR #528, opened 2025-04-10
fix: use conda env only when pipeline is comfyui

eliteprox:
This change runs the noop pipeline normally and ensures the comfyui pipeline is started in the comfystream environment

### DIFF
--- a/runner/app/pipelines/live_video_to_video.py
+++ b/runner/app/pipelines/live_video_to_video.py
@@ -107,10 +107,7 @@ class LiveVideoToVideoPipeline(Pipeline):
             raise ConnectionError(f"Failed to get status: {e}")
 
     def start_process(self, **kwargs):
-        if kwargs.get("pipeline") == "comfyui":
-            cmd = ["conda", "run", "--no-capture-output", "-n", "comfystream", "python", str(self.infer_script_path)]
-        else:
-            cmd = ["python", str(self.infer_script_path)]
+        cmd = [sys.executable, str(self.infer_script_path)]
 
         # Add any additional kwargs as command-line arguments
         for key, value in kwargs.items():
--- a/runner/app/pipelines/live_video_to_video.py
+++ b/runner/app/pipelines/live_video_to_video.py
@@ -110,7 +110,7 @@ class LiveVideoToVideoPipeline(Pipeline):
         if kwargs.get("pipeline") == "comfyui":
             cmd = ["conda", "run", "--no-capture-output", "-n", "comfystream", "python", str(self.infer_script_path)]
         else:
-            cmd = [str(self.infer_script_path)]
+            cmd = ["python", str(self.infer_script_path)]
 
         # Add any additional kwargs as command-line arguments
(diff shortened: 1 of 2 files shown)


runner/app/pipelines/live_video_to_video.py
```diff
@@ -107,7 +107,10 @@ class PipelineStatus(BaseModel):
             raise ConnectionError(f"Failed to get status: {e}")
 
     def start_process(self, **kwargs):
-        cmd = [str("/workspace/miniconda3/envs/comfystream/bin/python"), str(self.infer_script_path)]
+        if kwargs.get("pipeline") == "comfyui":
+            cmd = ["conda", "run", "--no-capture-output", "-n", "comfystream", "python", str(self.infer_script_path)]
+        else:
+            cmd = [str(self.infer_script_path)]
```
j0sh, line 113: Does this work without `python` as the first element? Before the #479 it was

https://github.com/livepeer/ai-runner/blob/aa1700685b3a4ef3e77bbbf046fa7cf97c407285/runner/app/pipelines/live_video_to_video.py#L110

eliteprox: Probably not, let me check that again

eliteprox: Rebased, fixed and tested noop pipeline.

leszko: I changed to use "python". Otherwise it may not work if you don't have correct file permissions. In my case it was missing "x", so it didn't work.

victorges (approved):

leszko (approved):

runner/app/pipelines/live_video_to_video.py
```diff
@@ -107,7 +107,10 @@ class PipelineStatus(BaseModel):
             raise ConnectionError(f"Failed to get status: {e}")
 
     def start_process(self, **kwargs):
-        cmd = [str("/workspace/miniconda3/envs/comfystream/bin/python"), str(self.infer_script_path)]
+        if kwargs.get("pipeline") == "comfyui":
+            cmd = ["conda", "run", "--no-capture-output", "-n", "comfystream", "python", str(self.infer_script_path)]
```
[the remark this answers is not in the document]
eliteprox: I'll give this a try. This may work only due to the parent process `uvicorn` originating from the comfystream env: https://github.com/livepeer/ai-runner/blob/abd6e3293c7d23b07bffc8cdca9a082c632524a4/runner/docker/Dockerfile.live-app__PIPELINE__#L37

Here's some more info on what's happening. There are two python environments in the image: 
- Base env is used by default: `/workspace/miniconda3/bin/python` (dl_checkpoints.sh is trying to run in here, missing packages)
- Comfystream conda env at `/workspace/miniconda3/envs/comfystream/bin/python` - has correct packages for dl-checkpoints, but is not activated on non-interactive shell commands any longer. 

We might begin using the base env instead of creating a separate one for comfystream to help prevent this

PR #530, opened 2025-04-10
add depth-anything-2-large compilation to dl_checkpoints

eliteprox:
This change builds an additional depth-anything engine using the new model `depth_anything_v2_vitl.onnx` added in https://github.com/livepeer/comfystream/pull/79

### DIFF
--- a/runner/dl_checkpoints.sh
+++ b/runner/dl_checkpoints.sh
@@ -124,8 +124,8 @@ function build_tensorrt_models() {
   # Depth-Anything-Tensorrt
   docker run --rm -v ./models:/models --gpus all -l TensorRT-engines $AI_RUNNER_COMFYUI_IMAGE \
     bash -c "cd /workspace/ComfyUI/models/tensorrt/depth-anything && \
-                $CONDA_PYTHON /workspace/ComfyUI/custom_nodes/ComfyUI-Depth-Anything-Tensorrt/export_trt.py --trt_path=./depth_anything_v2_vitl-fp16.engine --onnx-path=./depth_anything_v2_vitl.onnx && \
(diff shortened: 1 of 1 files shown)


runner/dl_checkpoints.sh
```diff
@@ -124,7 +124,8 @@ function build_tensorrt_models() {
   # Depth-Anything-Tensorrt
   docker run --rm -v ./models:/models --gpus all -l TensorRT-engines $AI_RUNNER_COMFYUI_IMAGE \
     bash -c "cd /workspace/ComfyUI/models/tensorrt/depth-anything && \
-                $CONDA_PYTHON /workspace/ComfyUI/custom_nodes/ComfyUI-Depth-Anything-Tensorrt/export_trt.py && \
+                $CONDA_PYTHON /workspace/ComfyUI/custom_nodes/ComfyUI-Depth-Anything-Tensorrt/export_trt.py --trt_path=./depth_anything_v2_vitl-fp16.engine --onnx-path=./depth_anything_v2_vitl.onnx && \
```
pschroedl, line 127: There is an inconsistency here between `--onnx-path` on line 127 and `--onnx_path` on line 128 - I'm assuming the latter is correct.

pschroedl (changes requested):

runner/dl_checkpoints.sh
eliteprox: These filenames appear to be correct. This is due to two depth-anything models being built now instead of one. In `models.yaml` these are the correct onnx files:

- [depth_anything_vitl14.onnx](https://github.com/livepeer/comfystream/blob/22c60af78715badc5db1447e3b0ccf9f43446d31/configs/models.yaml#L23)

- [depth_anything_v2_vitl.onnx](https://github.com/livepeer/comfystream/blob/22c60af78715badc5db1447e3b0ccf9f43446d31/configs/models.yaml#L27)

Both are built in [entrypoint.sh](https://github.com/livepeer/comfystream/blob/22c60af78715badc5db1447e3b0ccf9f43446d31/docker/entrypoint.sh#L70-L78) for the comfystream demo image

victorges (approved):

PR #531, opened 2025-04-11
Add missing logging import to last-value-cache

j0sh:
Fixes an exception ... don't think this has happened in prod but I managed to trigger it somehow in dev.


j0sh:
update: it happens in prod now

leszko (approved):

mjh1 (approved):

PR #534, opened 2025-04-14
Fix `dl_checkpoints.sh`

stronk-dev:
[The script that gets downloaded](https://raw.githubusercontent.com/yondonfu/comfystream/535c71a6f665bc1169d07cddd4e2b3cf4edd5a82/src/comfystream/scripts/build_trt.py) does not expose the new parameters.
However, the PR that puts the correct script in the image [has been merged](https://github.com/livepeer/comfystream/blob/22c60af78715badc5db1447e3b0ccf9f43446d31/src/comfystream/scripts/build_trt.py#L86-L90). After pulling the latest version, I was able to run the script with the min/max width/height parameters.

### DIFF
--- a/runner/dl_checkpoints.sh
+++ b/runner/dl_checkpoints.sh
@@ -146,12 +146,9 @@ function build_tensorrt_models() {
     )
   
   # Dreamshaper-8-Dmd-1kstep static dynamic 488x704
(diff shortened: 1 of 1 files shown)


eliteprox (approved):
LGTM! Thanks for the fix

PR #535, opened 2025-04-14
wip: refactor to use comfystream.pipeline

eliteprox:
This change makes use of the exported pipeline class from https://github.com/livepeer/comfystream/pull/114


runner/app/live/streamer/process.py
```diff
@@ -109,7 +109,7 @@ def get_recent_logs(self, n=None) -> list[str]:
 
     def process_loop(self):
         self._setup_logging()
-        pipeline = None
```
eliteprox, line 112: Should be reverted

runner/app/live/streamer/process.py
```diff
@@ -109,7 +109,7 @@ def get_recent_logs(self, n=None) -> list[str]:
...
+#        pipeline = None
```
eliteprox, line 112: Note to revert later

runner/app/live/pipelines/comfyui.py
```diff
@@ -52,52 +52,62 @@ def validate_prompt(cls, v) -> dict:
...
-        frame.side_data.input = torch.tensor(image_np).unsqueeze(0)
-        frame.side_data.skipped = True
-        self.client.put_video_input(frame)
-        await self.video_incoming_frames.put(VideoOutput(frame, request_id))
-
-    async def get_processed_video_frame(self):
-        result_tensor = await self.client.get_video_output()
-        out = await self.video_incoming_frames.get()
-        while out.frame.side_data.skipped:
-            out = await self.video_incoming_frames.get()
```
victorges, line 86: Don't we need this logic anymore?

victorges (changes requested):
Neat! Requesting changes though because there's a regression on the handling of the request_id. It's basically rolling back https://github.com/livepeer/ai-runner/pull/525

runner/app/live/pipelines/comfyui.py
eliteprox: > Don't we need this logic anymore?

The frame skipping logic is in comfystream here https://github.com/livepeer/comfystream/blob/e9b87bc7bc0adb32c5731f70b1e11e44c1f2e3c1/src/comfystream/pipeline.py#L169-L170

I also fixed the `request_id` side_data passthrough

runner/app/live/pipelines/interface.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, **params):
         pass
 
     @abstractmethod
-    async def put_video_frame(self, frame: VideoFrame, request_id: str):
+    async def put_video_frame(self, frame: VideoFrame):
```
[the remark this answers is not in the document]
eliteprox: Thanks for pointing out the solution here, I have opened a PR to comfystream that will support passing an optional request_id into the frame side_data. Tested this solution successfully

PR #541, opened 2025-04-24
Retry decoder if it throws an exception.

j0sh:
This isn't the most elegant approach since there may still be junk left over in the pipe but the ffmpeg mpegts demuxer will sync up correctly both to a start marker and an IDR, provided that the keyframe interval isn't too long.

### DIFF
--- a/runner/app/live/trickle/media.py
+++ b/runner/app/live/trickle/media.py
@@ -11,6 +11,9 @@ from .trickle_publisher import TricklePublisher
 from .decoder import decode_av
 from .encoder import encode_av
 
+MAX_DECODER_RETRIES = 3
+DECODER_RETRY_RESET_SECONDS = 120 # reset retry counter after 2 minutes
+
 MAX_ENCODER_RETRIES = 3
 ENCODER_RETRY_RESET_SECONDS = 120 # reset retry counter after 2 minutes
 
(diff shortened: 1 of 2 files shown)


mjh1 (approved):

j0sh:
Added a few more small tweaks in https://github.com/livepeer/ai-runner/pull/541/commits/a0989b9fedd9e92c4c8417c3e000b9d377d1db7d - from the commit message

Improve exception handling in run_subscribe
 
Ignore errors from start-up; usually indicates stream terminated
before enough data came in to initialize things.

runner/app/live/trickle/media.py
```diff
@@ -26,8 +26,7 @@ async def run_subscribe(subscribe_url: str, image_callback, put_metadata, monito
         await asyncio.gather(subscribe_task, parse_task)
         logging.info("run_subscribe complete")
     except Exception as e:
-        logging.error(f"preprocess got error {e}", e)
-        raise e
+        logging.exception("run_subscribe got error", stack_info=True)
```
mjh1, line 29: why don't we want to print the actual error `e`?

mjh1 (approved):

runner/app/live/trickle/media.py
j0sh: [logging.exception](https://docs.python.org/3/library/logging.html#logging.Logger.exception) will print it out, its magic

PR #543, opened 2025-04-28
Update default workflow for depth-anything-2

eliteprox:


eliteprox:
For this change, we are waiting for prd approval. There is an FPS reduction of approx 2-3

eliteprox:
We still need to deploy the new depth-anything model to primary staging before merging

PR #546, opened 2025-05-01
fix: comfyui pipeline shutdown logic

eliteprox:
This change resolves an error when shutting down the comfyui pipeline via `_restart_process` in [process_guardian.py](https://github.com/livepeer/ai-runner/blob/21c5539a014f32c7371892b9ae53bf81fb77821e/runner/app/live/streamer/process_guardian.py#L173) due to the various i/o loops started by `run_pipeline_loops` within `PipelineProcess` not stopping gracefully with the pipeline. 

`_run_pipeline_loops` has been updated to include proper error handling and stop all running tasks when any of them are cancelled or stopped.

`pipeline` is also updated to `self.pipeline` instead of passing the variable to functions in the same class

### DIFF
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -172,10 +172,10 @@ class PipelineProcess:
                     raise
 
     async def _run_pipeline_loops(self):
-        pipeline = await self._initialize_pipeline()
-        input_task = asyncio.create_task(self._input_loop(pipeline))
-        output_task = asyncio.create_task(self._output_loop(pipeline))
-        param_task = asyncio.create_task(self._param_update_loop(pipeline))
+        self.pipeline = await self._initialize_pipeline()
+        input_task = asyncio.create_task(self._input_loop())
+        output_task = asyncio.create_task(self._output_loop())
+        param_task = asyncio.create_task(self._param_update_loop())
 
         async def wait_for_stop():
             while not self.is_done():
@@ -225,21 +225,22 @@ class PipelineProcess:
             raise  # Re-raise to ensure the error is propagated
         finally:
             try:
-                await pipeline.stop()
+                if self.pipeline:
+                    await self.pipeline.stop()
             except Exception as e:
                 self._report_error(f"Error stopping pipeline: {str(e)}")
             finally:
-                pipeline = None
+                self.pipeline = None
                 
             logging.info("Successfully stopped pipeline")
 
-    async def _input_loop(self, pipeline: Pipeline):
+    async def _input_loop(self):
         while not self.is_done():
             try:
                 input_frame = await asyncio.to_thread(self.input_queue.get)
                 if isinstance(input_frame, VideoFrame):
                     input_frame.log_timestamps["pre_process_frame"] = time.time()
-                    await pipeline.put_video_frame(input_frame, self.request_id)
+                    await self.pipeline.put_video_frame(input_frame, self.request_id)
                 elif isinstance(input_frame, AudioFrame):
                     await asyncio.to_thread(self.output_queue.put, AudioOutput([input_frame], self.request_id))
             except queue.Empty:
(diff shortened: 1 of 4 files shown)


runner/app/live/streamer/process.py
```diff
@@ -172,57 +172,86 @@ async def _initialize_pipeline(self):
                     raise
 
     async def _run_pipeline_loops(self):
-        pipeline = await self._initialize_pipeline()
-        input_task = asyncio.create_task(self._input_loop(pipeline))
-        output_task = asyncio.create_task(self._output_loop(pipeline))
-        param_task = asyncio.create_task(self._param_update_loop(pipeline))
+        self.pipeline = await self._initialize_pipeline()
```
victorges, line 175: The main reason for not using fields instead of args is that it can be a little tricky. This class starts a new process so then it becomes two separate objects, one in the original class and one in the subprocess. It can be error-prone to rely on fields and then have things not really updated.

So if this is not a hard requirement, I'd say we should avoid fields as much as possible or consider creating a whole new class which is actually the "process state", which we only access from within the sub-process. WDYT?

runner/app/live/streamer/process.py
```diff
@@ -172,57 +172,86 @@ async def _initialize_pipeline(self):
...
+            
+            logging.info("PipelineProcess: Pipeline shutting down") 
+            
+            # Check if any task failed and get the error details
+            results = await asyncio.gather(*done, return_exceptions=True)
+            for task, result in zip(done, results):
+                if isinstance(result, Exception):
+                    task_name = task.get_name() if task.get_name() else 'unknown'
+                    self._report_error(f"Error while cancelling {task_name} task: {str(result)}")
+                    self.done.set()
```
victorges, line 199: We should always stop all tasks by setting done. There's no case where we want to keep running just a part of the tasks but keep the other ones running. So this could probably be done around line 190. WDYT?

runner/app/live/streamer/process.py
```diff
@@ -172,57 +172,86 @@ async def _initialize_pipeline(self):
...
+            
+            # Cancel remaining tasks and gather results
+            if pending:
+                results = await asyncio.gather(*pending, return_exceptions=True)
```
victorges, line 203: We need a timeout here, or maybe cancel all `pending` tasks before doing this gather. Otherwise this will run forever if any of the tasks doesn't finish gracefully.

runner/app/live/streamer/process.py
```diff
@@ -172,57 +172,86 @@ async def _initialize_pipeline(self):
...
+                for task, result in zip(pending, results):
+                    if isinstance(result, Exception):
+                        task_name = task.get_name() if task.get_name() else 'unknown'
+                        self._report_error(f"Error while cancelling {task_name} task: {str(result)}")
+                        self.done.set()
```
victorges, line 208: This doesn't seem needed here. As stated in the other comment, we should flag `done` whenever _any_ task fails so that the rest of the tasks start their cleanup.

runner/app/live/streamer/process.py
```diff
@@ -172,57 +172,86 @@ async def _initialize_pipeline(self):
...
+                
         except Exception as e:
-            self._report_error(f"Error in pipeline loops: {e}")
+            self._report_error(f"Fatal error in pipeline loops: {str(e)}")
+            self.done.set()
+            raise  # Re-raise to ensure the error is propagated
```
victorges, line 213: No one will catch this, this is the main function ran by the process, and we already have the log above (which should be critical if we ever see in production). I think it's better to remove the `raise` to avoid any confusion.

runner/app/live/streamer/process.py
```diff
@@ -172,57 +172,86 @@ async def _initialize_pipeline(self):
...
         async def wait_for_stop():
             while not self.is_done():
                 await asyncio.sleep(0.1)
-
-        tasks = [input_task, output_task, param_task, asyncio.create_task(wait_for_stop())]
+            logging.info("PipelineProcess: Pipeline loops interrupted, stopping pipeline")
 
         try:
-            await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+            self.tasks = [input_task, output_task, param_task, asyncio.create_task(wait_for_stop())]
```
victorges, line 186: Doesn't seem like this is used anywhere else, so it should be a regular variable and not a field. As stated in the other comment, we should generally avoid fields in this class as they can be misleading and error-prone.

victorges (commented):
Some general impressions on the current proposed flow

PR #547, opened 2025-05-05
trickle: Start subscribing at the current seq, rather than the next

j0sh:
This helps mitigate a bit of delay if the runner starts subscribing to a channel after the first segment has already started writing. This would have made the runner start waiting for the *next* segment (1s on average, up to 2s), which seems to happen roughly 2/3s of the time by eyeballing this histogram. Requires https://github.com/livepeer/go-livepeer/pull/3552

image

### DIFF
--- a/runner/app/live/trickle/trickle_subscriber.py
+++ b/runner/app/live/trickle/trickle_subscriber.py
@@ -4,9 +4,9 @@ import logging
 import sys
 
 class TrickleSubscriber:
-    def __init__(self, url: str, max_retries=5):
+    def __init__(self, url: str, start_seq=-2, max_retries=5):
         self.base_url = url
-        self.idx = -1  # Start with -1 for 'latest' index
+        self.idx = start_seq
         self.pending_get: aiohttp.ClientResponse | None = None  # Pre-initialized GET request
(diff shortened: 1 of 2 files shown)


leszko (approved):

runner/app/live/trickle/trickle_subscriber.py
```diff
@@ -4,9 +4,9 @@
 import sys
 
 class TrickleSubscriber:
-    def __init__(self, url: str, max_retries=5):
+    def __init__(self, url: str, start_seq=-2, max_retries=5):
```
victorges, line 7: Is this feature already implemented on the O side? Does `-2` mean getting the last "complete" segment and not the current (next) one being pushed? Or will `-1` pick the next seq even if it hasn't started yet?

victorges (approved):
LGTM

runner/app/live/trickle/trickle_subscriber.py
victorges: nvm just found https://github.com/livepeer/go-livepeer/pull/3552/files

mjh1 (approved):

PR #548, opened 2025-05-06
Shut down job if decoder stops.

j0sh:
Previously stopping the decoder would not stop any other processing except for blocking trickle subscriptions (due to the pipe being full) and the runner would be waiting until inference times out after 60 seconds of no input.

Here we fix this by signaling two things:

* Shut down the trickle subscriber by closing the end of the file descriptor that it writes into. This in turn ends the run_subscribe function.

* Send an empty input frame to inference to force the rest of the process to shut down, because run_subscribe stopping on its own apparently does not trigger any other teardown.

### DIFF
--- a/runner/app/live/trickle/media.py
+++ b/runner/app/live/trickle/media.py
@@ -4,7 +4,6 @@ import asyncio
 import logging
 import os
 import threading
-import subprocess
 
 from .trickle_subscriber import TrickleSubscriber
 from .trickle_publisher import TricklePublisher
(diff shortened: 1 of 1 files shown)


leszko (approved):

runner/app/live/trickle/media.py
```diff
@@ -97,6 +98,8 @@ def decode_runner():
                     logging.exception(f"Error in decode_av, retrying {retry_count}/{MAX_DECODER_RETRIES}", stack_info=True)
                 else:
                     logging.exception("Error in decode_av, maximum retries reached", stack_info=True)
+                    # force write end of pipe to close to terminate trickle subscriber
+                    write_fd.close()
```
victorges, line 102: Maybe move this to outside the loop, like close to `os.close(in_pipe)` in case we exit through some other path? There's a `break` on L89 👀 

Or even better if all that clean up would be in a finally block, but we'd need to indent the whole `while` loop here 😫

victorges (approved):
LGTM anyway

runner/app/live/trickle/media.py
j0sh: I didn't really want to change the behavior too much since IIRC we'd get _another_ asyncio exception if trying to close an already-closed FD which would be the typical behavior under most normal shutdowns. So we'd need nested exception handling in addition to the `finally`, which is also annoying.

I agree it is kind of fragile if another exit path slips through in the future. Moving closer to the `in_pipe` is an option.

> There's a break on L89 👀

Good spot, the `write_fd` would already be closed in this case. That just prevents us from retrying on an input that we already know to be closed.

victorges: sounds good let's keep this minimal to include it in the deploy tmr

mjh1 (approved):

PR #550, opened 2025-05-06
live/streamer: Quick stream failover on process restart

victorges:
This is to make sure we (the gateway/app) quickly failover to a different runner when the inference
process crashes. Otherwise they would have to wait for at least 30s until it comes back.

The risk of this change is in case someone finds out a prompt that always crashes the process. This 
would make it a kryptonite that could take our whole infra if the gateway/app retries indefinitely 💥 

We haven't seen any of that until now + if we have a reasonable number of app/gateway retries this
could be somewhat safe to merge anyway. 

This is built on top of #549 but is unrelated. Only implemented together but decided to put out for review separately.

### DIFF
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -42,9 +42,6 @@ class PipelineProcess:
         return self.process.is_alive()
 
     async def stop(self):
-        self._stop_sync()
-
-    def _stop_sync(self):
         self.done.set()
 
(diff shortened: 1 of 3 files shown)


j0sh (approved):

leszko (approved):

victorges:
ok finally managed to test this on mac box. still without my gpu dev server :(

PR #551, opened 2025-05-07
Catch all exceptions from trickle subscribe preconnects.

j0sh:
Sometimes we get Timeouts (default ~5 minutes) if for some reason the orchestrator does not close the trickle channel.

This would be an uncaught exception that crashes the process so just catch everything here to avoid any crashes.

### DIFF
--- a/runner/app/live/trickle/trickle_subscriber.py
+++ b/runner/app/live/trickle/trickle_subscriber.py
@@ -56,8 +56,8 @@ class TrickleSubscriber:
                 resp.release()
                 logging.error(f"Trickle sub Failed GET {url} status code: {resp.status}, msg: {body}")
 
-            except aiohttp.ClientError as e:
-                logging.error(f"Trickle sub Failed to complete GET {url} error: {e}")
+            except Exception:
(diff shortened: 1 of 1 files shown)


emranemran (approved):

leszko (approved):

mjh1 (approved):

runner/app/live/trickle/trickle_subscriber.py
```diff
@@ -56,8 +56,8 @@ async def preconnect(self):
                 resp.release()
                 logging.error(f"Trickle sub Failed GET {url} status code: {resp.status}, msg: {body}")
 
-            except aiohttp.ClientError as e:
-                logging.error(f"Trickle sub Failed to complete GET {url} error: {e}")
+            except Exception:
+                logging.exception(f"Trickle sub Failed to complete GET {url}", stack_info=True)
```
victorges, line 60: Wow I thought `stack_info` was precluded with `logging.exception`. Apparently not from func signature at least 🧐

victorges (approved):

PR #564, opened 2025-05-12
live/trickle: Avoid downscaling input frames more than necessary

victorges:
We are currently resizing the input frames to 512 pixels, but we set the _lowest_ dimension
to 512. This means that we will downscale further than necessary, then crop and _upscale_
back the image on [the `streamer` side.](https://github.com/livepeer/ai-worker/blob/d06894afd00ef16eb3a09b296d56866d146f3c9d/runner/app/live/streamer/streamer.py#L196)

I tried looking around for a while to see if there was a way to crop the `VideoFrame` from within
pyav, but apparently there is not. The solution would be converting the frame to a numpy array
and cropping it there, which is exactly what we are doing on the streamer side anyway. To avoid 
duplication, I kept only the resize on the pyav decode logic, and only made sure we  don't call a
noop cv2 resize on the streamer side in case it was already resized to the right size.

If we want to go further, I think we could centralize all this frame manipulation logic on either the
decode code or the streamer code. I'll leave that for the future tho (e.g. when we want this cropping
to be parametrized).

### DIFF
--- a/runner/app/live/streamer/streamer.py
+++ b/runner/app/live/streamer/streamer.py
@@ -178,14 +178,14 @@ class PipelineStreamer(ProcessCallbacks):
             if frame.mode != "RGBA":
                 frame = frame.convert("RGBA")
 
-            # crop the max square from the center of the image and scale to 512x512
-            # most models expect this size especially when using tensorrt
+            # Scale image to 512x512 as most models expect this size, especially when using tensorrt
             width, height = frame.size
             if (width, height) != (512, 512):
                 frame_array = np.array(frame)
 
+                # Crop to the center square if image not already square
+                square_size = min(width, height)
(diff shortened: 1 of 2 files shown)


j0sh (approved):
Looks okay, this would come at the expense of shuffling a bit more data around but it is probably manageable. FWIW we had trouble sending around full size images, otherwise we wouldn't bother resizing at all.

> I tried looking around for a while to see if there was a way to crop the VideoFrame from within pyav, but apparently there is not.

That is not surprising. The way to crop + resize in a single pass is a bit of a trick: set a starting offset into the underlying buffer and your desired width / height while keeping the underlying linesize the same. We may be able to manually construct a VideoFrame to do that but I haven't tried.

PR #566, opened 2025-05-12
.github: Fix docker image build on PRs

victorges:
Since we skip the base image build when unchanged, we can't always set the docker digest on the base image. So only specify if it is present.


victorges:
successful build is enough to merge this. rebasing and merging.

hjpotter92:
with the `needs:` clause in workflow, the runner builds will also not be triggered unless base image was built.

PR #568, opened 2025-05-12
live/streamer: Improve monitor loop reliability

victorges:
This is to rewrite the monitor logic in the process guardian/streamer both to simplify it 
as well as make it more reliable.

This implements INF-119

## Previous implementation
- Independent logic on process guardian to restart process and streamer to declare state
- Passive, state was only computed/updated when requested on healthcheck
- All the other problems described in the issue

## New implementation
- Centralized state tracking on a single place: process guardian monitor loop
- It runs every 1s and computes the current state
- When the state is ERROR, it tries restarting the process up to 3 times
   - ERROR state is sticky, until process gets restarted or container is killed
- State is updated on the status accordingly
- Shorter timeouts for errors:
   - 5s not producing output frames (after being active)
   - 30s not loading the pipeline after a simple params update
   - 120s not loading the pipeline on startup (was 240s before)

Also cleaned up other stuff along the way, like the `stream_stopped` function which can
be a `is_stream_running` callback instead so we avoid the state replication. With the new
`StreamCallbacks`, we can allow the process guardian to call the streamer for checking
state or triggering events (for now just a `stop`) with a well-defined interface to avoid 
circular references.

## Testing

Ran noop container on local box with virtual hang ups and ensure things were running smooth.

### DIFF
--- a/runner/app/live/streamer/process_guardian.py
+++ b/runner/app/live/streamer/process_guardian.py
@@ -6,7 +6,7 @@ import abc
 from trickle import InputFrame, OutputFrame
 
 from .process import PipelineProcess
-from .status import PipelineState, PipelineStatus
+from .status import PipelineState, PipelineStatus, InferenceStatus
 
 
 class StreamerCallbacks(abc.ABC):
@@ -22,7 +22,12 @@ class StreamerCallbacks(abc.ABC):
     async def emit_monitoring_event(self, event_data: dict) -> None: ...
 
     @abc.abstractmethod
-    def trigger_stop_stream(self) -> bool: ...
+    def trigger_stop_stream(self) -> bool:
+        """
+        Trigger a stop of the stream. Returns True if the stream was actually
+        stopped on this call or False if it was already stopped.
+        """
+        ...
 
 
 class ProcessGuardian:
@@ -37,7 +42,7 @@ class ProcessGuardian:
         params: dict,
     ):
         self.pipeline = pipeline
-        self.params = params
+        self.initial_params = params
         self.streamer: StreamerCallbacks = _NoopStreamerCallbacks()
 
         self.process: Optional[PipelineProcess] = None
@@ -47,8 +52,8 @@ class ProcessGuardian:
         )
 
     async def start(self):
-        self.process = PipelineProcess.start(self.pipeline, self.params)
-        self.status.state = PipelineState.INITIALIZING
+        self.process = PipelineProcess.start(self.pipeline, self.initial_params)
+        self.status.update_state(PipelineState.INITIALIZING)
         self.monitor_task = asyncio.create_task(self._monitor_loop())
 
     async def stop(self):
@@ -112,10 +117,7 @@ class ProcessGuardian:
     async def update_params(self, params: dict):
         if not self.process:
             raise RuntimeError("Process not running")
-        self.params = params
-        logging.info(
-            f"ProcessGuardian: Queuing parameter update with hash={self.status.inference_status.last_params_hash}, params={params}"
-        )
+
         self.process.update_params(params)
         self.status.update_params(params)
 
@@ -129,7 +131,7 @@ class ProcessGuardian:
             }
         )
         logging.info(
-            f"ProcessGuardian: Parameter update queued and monitoring callback completed. Hash={self.status.inference_status.last_params_hash}"
+            f"ProcessGuardian: Parameter update queued. hash={self.status.inference_status.last_params_hash} params={params}"
         )
 
     def get_status(self, clear_transient: bool = False) -> PipelineStatus:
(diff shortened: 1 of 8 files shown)


j0sh (approved):

runner/app/live/streamer/process_guardian.py
```diff
@@ -121,68 +130,92 @@ async def update_params(self, params: dict):
...
             self.status.inference_status.last_params = None
             self.status.inference_status.last_restart_logs = None
         return status
 
-    def _current_state(self) -> str:
-        # Hot fix: the comfyui pipeline process is having trouble shutting down and causes restarts not to recover.
-        # So return an error state if the process has restarted so the worker will restart the whole container.
-        # TODO: Remove this once pipeline shutodwn is fixed and restarting process is useful again.
-        if self.status.inference_status.restart_count > 0:
+    def _compute_current_state(self) -> str:
```
leszko, line 145: I think that after this PR, the health check stopped working. At least for the noop box setup. Could you check? Here's what I try:

In go-livepeer:
```
$ make box-orchestrator
...
2025/05/14 11:35:24 ERROR Container health check failed too many times container=live-video-to-video_noop_8900
2025/05/14 11:35:24 INFO Removing managed container gpu=emulated-0 name=live-video-to-video_noop_8900 modelID=noop
2025/05/14 11:35:35 INFO Container was kept warm, restarting container=live-video-to-video_noop_8900
2025/05/14 11:35:35 INFO Starting managed container gpu=emulated-0 name=live-video-to-video_noop_8900 modelID=noop containerImage=livepeer/ai-runner:live-app-noop
2025/05/14 11:35:37 INFO Watching container container=live-video-to-video_noop_8900
2025/05/14 11:35:42 ERROR Container not healthy container=live-video-to-video_noop_8900 status=ERROR failures=1
2025/05/14 11:35:47 ERROR Container not healthy container=live-video-to-video_noop_8900 status=ERROR failures=2
2025/05/14 11:35:52 ERROR Container not healthy container=live-video-to-video_noop_8900 status=ERROR failures=3
```

It works fine for `main`

runner/app/live/streamer/process_guardian.py
```diff
@@ -121,68 +130,92 @@ async def update_params(self, params: dict):
...
+        # Special case: process not running or initializing
+        if not self.process or not self.process.is_alive():
+            logging.error("Process is not alive. Returning ERROR state")
+            return PipelineState.ERROR
+        elif not self.process.is_pipeline_initialized() or self.process.done.is_set():
+            # done is only set in the middle of the restart process so also return INITIALIZING
```
leszko, line 151: Random question, is the runner health reported as `OK 200` when the comfyui is loaded and it's ready to serve the stream? Or is it reporting `OK 200` even before comfyui is loaded?

Asking, because I remember some time ago it was a problem that when Orchestrator started and it all of the sudden received the request then it was super slow, because it had to wait for loading the comfyui. I wonder if it's already fixed.

leszko (commented):
Added 2 comments.

runner/app/live/streamer/process_guardian.py
victorges: Fixed on the latest commits! I hadn't had a chance to run it yet so there were some dumb errors. I improved some other stuff I found out during the testing as well!

runner/app/live/streamer/process_guardian.py
victorges: With this change, the `infer.py` healthcheck will return `INITIALIZING` while the process is loading the pipeline. The runner API will then convert that to an `OK` [here](https://github.com/livepeer/ai-worker/blob/9e884264c3513097e717679a7c7b99ab133d69a9/runner/app/pipelines/live_video_to_video.py#L97), which actually means "busy" on the live containers on the [`watchContainer` logic](https://github.com/livepeer/go-livepeer/blob/e6a3bf4bb9da28daf8f0c1df70fad56326db9d4c/ai/worker/docker.go#L601).

We will need to do some refactoring though, because:
- `OK` is also the only state returned by [batch AI containers](https://github.com/livepeer/ai-worker/blob/b7aee366e74135710a102fe577f6e0880a3f90cf/runner/app/pipelines/base.py#L28). We can't simply change that to `IDLE` though because it will cause the worker not to wait for their requests to finish before opening it up for another request.
- The worker currently starts the containers as "idle" already. So we need some logic on `watchContainer` to automatically "phantom borrow" it when it sees the `INITIALIZING` state on the healthcheck, until it becomes `IDLE` again

So I think we'll need to propagate the `INITIALIZING` on that runner API code and then implement the worker logic to handle that. I'll implement these on separate PRs though to test thnings separately. This PR is fine as is, with the only difference that we'll be returning `OK` on the beginning instead of `IDLE`, which will be a noop for the worker anyway.

victorges: Simple one on runner API: https://github.com/livepeer/ai-runner/pull/576

victorges: And the worker `watchContainer` impl: https://github.com/livepeer/go-livepeer/pull/3569

leszko (approved):

victorges:
Merged https://github.com/livepeer/go-livepeer/pull/3569 and will wait for build to finish before merging this (otherwise LOADING state would be considered an error)

PR #572, opened 2025-05-14
Queue up audio to stay in sync with video

j0sh:
We drop audio while waiting for the first video frame, but often video is slightly delayed going through the ComfyUI pipeline. This causes us to more or less immediately drop all the audio around the first frame.

If enough audio is missing then this causes problems downstream when audio eventually shows up, because it appears very "late" compared to the first video frame (because we have dropped the earlier audio around the first frame).

Avoid this by queuing up some audio, up to a limit. If things seem too far out of sync (>2s) then just stop the stream.


runner/app/live/trickle/encoder.py
```diff
@@ -82,35 +96,92 @@ def custom_io_open(url: str, flags: int, options: dict):
             frame.pts = rescale_ts(avframe.timestamp, avframe.time_base, output_video_stream.codec_context.time_base)
             frame.time_base = output_video_stream.codec_context.time_base
             current = avframe.timestamp * avframe.time_base
+            if len(audio_buffer) > 0:
+                # check if video is too far behind audio. if so, drop to try and catch up
+                audio_ts = audio_buffer[0].timestamp * audio_buffer[0].time_base
+                if audio_ts - current > AUDIO_BUFFER_SECS:
```
victorges, line 102: Is it possible for the video to come before the audio as well? What would happen on that case?

runner/app/live/trickle/encoder.py
```diff
@@ -82,35 +96,92 @@ def custom_io_open(url: str, flags: int, options: dict):
...
+                logging.info(f"Flushing {len(audio_buffer)} audio frames from={first_ts} to={last_ts} dropped_audio={dropped_audio} dropped_video={dropped_video}")
+                audio_buffer.clear()
+
+            av_broken = False
             for af in avframe.frames:
                 af.log_timestamps["frame_end"] = time.time()
                 log_frame_timestamps("Audio", af)
+                if not audio_received and video_received:
+                    first_audio_ts = af.timestamp * af.time_base
+                    if abs(first_audio_ts - first_video_ts) > AUDIO_BUFFER_SECS:
```
victorges, line 168: So this is only for the first segments, not an ongoing check right? I wonder if we want to have an ongoing logic anyway in case they diverge after start (unless that's really unlikely)

victorges (approved):

runner/app/live/trickle/encoder.py
j0sh: Yep, that case is handled just before writing the audio - https://github.com/livepeer/ai-runner/pull/572/files#diff-58ccb9520686dccb238caf0d92a8d842e08674435226cf5aeaaaf023bd17573fR169-R178

TBH a better implementation would queue up _either_ video or audio if we were expecting both, but I don't think we really want the additional latency on the video side here.

runner/app/live/trickle/encoder.py
j0sh: Correct, because the problems we are seeing are near the beginning, around track initialization when tools (mediamtx, ffmpeg, etc) are trying to figure out which tracks are present.

In general clients seem to have decent timestamp sync; we're just more prone to hitting those sync issues at the beginning right now due to the async nature of the inference pipeline, and some client behaviors during startup exacerbate the problem (eg, sending exactly one video frame at the beginning, followed by just audio for a couple seconds).

There's also the fact that since https://github.com/livepeer/ai-runner/pull/547 we are generally pulling "old" data that is behind the leading edge, so we are doing more rapid catch-up, and discarding more frames without having the opportunity to bypass those weird initial segments from the client. So lots of contributing factors here.

BTW the implication here is that if inference takes a while to start outputting frames (~4 seconds) then we are probably going to kick off the stream since we can't be certain that we're just behind (in queuing terms) or genuinely out of sync.

An ongoing sync check would probably be useful someday, but for now we just depend on the ffmpeg muxer behavior, where it interleaves for us and will drop late packets.

victorges: Got it, thanks for the exaplanation1

emranemran (approved):

PR #579, opened 2025-05-15
live: Add manifest ID to openapi and logs

j0sh:
Helps filter logs by orchestrator once swapping is in.

### DIFF
--- a/runner/gateway.openapi.yaml
+++ b/runner/gateway.openapi.yaml
@@ -525,7 +525,8 @@ components:
     AudioResponse:
       properties:
         audio:
-          $ref: '#/components/schemas/MediaURL'
+          allOf:
+          - $ref: '#/components/schemas/MediaURL'
           description: The generated audio.
       type: object
       required:
@@ -826,7 +827,8 @@ components:
     HTTPError:
       properties:
         detail:
-          $ref: '#/components/schemas/APIError'
+          allOf:
+          - $ref: '#/components/schemas/APIError'
           description: Detailed error information.
       type: object
       required:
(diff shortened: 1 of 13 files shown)


runner/gateway.openapi.yaml
```diff
@@ -878,11 +876,9 @@ components:
           title: Finish Reason
           default: ''
         delta:
-          allOf:
-          - $ref: '#/components/schemas/LLMMessage'
+          $ref: '#/components/schemas/LLMMessage'
         message:
-          allOf:
-          - $ref: '#/components/schemas/LLMMessage'
+          $ref: '#/components/schemas/LLMMessage'
```
leszko, line 881: I think you need to revert this OpenAPI changes from `allOf`, because the CI is failing. Maybe you're using some different lib version for generating OpenAPI.

leszko (approved):
Added one comment, other than that LGTM

runner/gateway.openapi.yaml
j0sh: Ah yeah this is an [older syntax](https://stackoverflow.com/a/73490368). Just tried a bunch of things to get CI to pick up the newer syntax (updated fastapi, updated python) but nothing seemed to work so I just manually picked the changes in https://github.com/livepeer/ai-runner/pull/579/commits/0aaed81f4ade646083dd2ef491b028c36357fd9a

> In OpenAPI 3.1 (which uses JSON Schema 2020-12 by default), there's even no need for allOf if you only have one $ref because $ref now [allows sibling keywords](https://stackoverflow.com/a/41752575/113116).

victorges (approved):

j0sh:
Holding off on merging this until after https://github.com/livepeer/ai-runner/pull/568

mjh1 (approved):

j0sh:
Rebased onto master and fixed merge conflicts, no functional changes.

Had to update the OpenAPI yamls to make CI happy since they weren't regenerated in  https://github.com/livepeer/ai-runner/pull/568 . This also probably needs a go-livepeer update (although I suspect it is not a functional change) - https://github.com/livepeer/go-livepeer/pull/3583

PR #580, opened 2025-05-15
live/streamer: Calculate FPS using a rolling window

victorges:
Fix the FPS calculation once and for all.

### DIFF
--- a/runner/app/live/streamer/process_guardian.py
+++ b/runner/app/live/streamer/process_guardian.py
@@ -3,6 +3,7 @@ import logging
 import time
 from typing import Optional
 import abc
+from collections import deque
 from trickle import InputFrame, OutputFrame
 
 from .process import PipelineProcess
@@ -51,6 +52,9 @@ class ProcessGuardian:
             params, False
         )
 
+        self.input_timestamps: deque[float] = deque()
+        self.output_timestamps: deque[float] = deque()
+
     async def start(self):
         self.process = PipelineProcess.start(self.pipeline, self.initial_params)
         self.status.update_state(PipelineState.LOADING)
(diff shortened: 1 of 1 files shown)


runner/app/live/streamer/process_guardian.py
```diff
@@ -3,9 +3,8 @@
 import time
 from typing import Optional
 import abc
-from collections import deque
-from trickle import InputFrame, OutputFrame
 
+from ..trickle import InputFrame, OutputFrame
```
j0sh, line 7: I don't Python so good, this is new syntax to me

j0sh (approved):
LGTM!

runner/app/live/streamer/process_guardian.py
victorges: This is making a relative import instead of an absolute one that relies on monkey patching the `syspath` to load the modules under `live` folder as absolute [here](https://github.com/livepeer/ai-worker/blob/05d6d6ff00cfa44938321f4a78ca9f0085b35590/runner/app/live/infer.py#L14-L16).

This is really annoying though because VS Code doesn't recognize the monkey patch so it keeps erroring on those imports. I'll start switching to relative imports wherever I can and at some point we may be able to enable a linter 😋

victorges: lol I actually had to revert this 🥲 
```
Traceback (most recent call last):
  File "/app/app/live/infer.py", line 12, in <module>
    from streamer import PipelineStreamer, ProcessGuardian
  File "/app/app/live/streamer/__init__.py", line 1, in <module>
    from .streamer import PipelineStreamer
  File "/app/app/live/streamer/streamer.py", line 12, in <module>
    from .process_guardian import ProcessGuardian, StreamerCallbacks
  File "/app/app/live/streamer/process_guardian.py", line 7, in <module>
    from ..trickle import InputFrame, OutputFrame
ImportError: attempted relative import beyond top-level package
```

PR #583, opened 2025-05-17
streamer: Fix pipeline process shutdown 

victorges:
This is to fix the process shutdown which was always needing a SIGKILL resulting in
shutdown issues even with the `noop` pipeline. The problem was that we were using
`asyncio.to_thread` without any timeout on the queue functions we call, resulting in
non-daemon threads being created and not getting shut down on tasks cancel. This
fixes it by adding a timeout to make sure the threads eventually stop by themselves.

There might still be some hanging threads/tasks on the `comfyui` pipeline. I'll test this
during the week at some point. If there are still issues, I'll revert the commit on `process_guardian`
that reenabled the restart logic and will merge just the fix to the `process` logic. If comfyui
got fixed as well then it's a huge win.

Gemini also had a message for you:
> To the reviewers: After an epic vibe-coding session, much head-scratching, and a joint effort in taming a host of elusive non-daemon threads, PipelineProcess has found its inner peace and now exits with the grace of a seasoned monk. This PR is the fruit of that collaborative debugging journey. Merge with tranquility!

### DIFF
--- a/runner/app/live/pipelines/noop.py
+++ b/runner/app/live/pipelines/noop.py
@@ -1,7 +1,7 @@
 import logging
 import asyncio
-import time
-import os
+from PIL import Image
+
 
 from .interface import Pipeline
 from trickle import VideoFrame, VideoOutput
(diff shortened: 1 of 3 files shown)


victorges:
@eliteprox FTR the pipeline shutdown is still not working with comfyui, so there is likely some thread leaking over there as well. Could you share an update on how that investigation is going? You'll likely be able to go further with it after this gets merged, since you'll actually get a clean shutdown now once comfystream side is fixed.

eliteprox (approved):
This LGTM, it makes sense to use the fifo queue for the audio passthrough. I'm not sure it will solve the shutdown process though

eliteprox:
> @eliteprox FTR the pipeline shutdown is still not working with comfyui, so there is likely some thread leaking over there as well. Could you share an update on how that investigation is going? You'll likely be able to go further with it after this gets merged, since you'll actually get a clean shutdown now once comfystream side is fixed.

I haven't had time to work on it in about a week until yesterday. I have confirmed that removing `_run_pipeline_loops` entirely will resolve it. The ComfyUI pipeline seems to always shutdown fine along with ComfyStreamClient. Your previous fix to exit the PipelineProcess entirely is sufficient for now, I'll should have a solution by end of this week and be able to finish https://github.com/livepeer/ai-runner/pull/546, however, feel free to take this on in the meantime if it is blocking anything (bit of an async pun there😄)

victorges:
@eliteprox 
> it makes sense to use the fifo queue for the audio passthrough. I'm not sure it will solve the shutdown process though

The FIFO queue is just to avoid the raw `put` without a timeout. We could as well add a timeout instead. In practice it doesn't make a lot of difference since these queues are not limited, so the `put` will work immediately and they won't block. This is just to be future-proof in case a limit is ever added, and is not the fix in this PR. The fix is adding a timeout on the `get` side, which were the threads that never got cancelled.

> I also find it interesting that infer.py is running continues to log between the pipeline shutdown starting and the timeout occurring. Could the issue be related to something related to streamer?

The `infer.py` process is not the pipeline (comfystream) process. It is the process that manages the pipeline process instead and runs the streamer. The `noop` shutdown is working fine now, so I'm pretty certain the only problem is with `comfystream` pipeline now, which is likely leaking an async task/thread similarly to how the `process.py` was doing before. I'd suggest you to take a look on the "vibe coding" commit in this PR to se all the debugging stuff that was added. It was even printing the names of the threads that were hanging the pipeline shutdown for example (but Gemini didn't actually figure out what was hanging, I had the insight from the `to_thread`).

PR #586, opened 2025-05-20
feat: custom resolution support for comfyui live pipeline

eliteprox:
Adds support for setting the input/output resolution of each stream with the pipeline parameters `width`, `height`:
```
    "params": {
        "width": 448,
        "height": 704,
        }
```

A couple areas to improve:
- Updates the default workflow to use a dynamic engine with 448 and 704 Latent image. This node must be dynamically updated to match the provided resolution of the request
- ComfyUI pipeline is still warmed at 512x512
- The input resolution is no longer scaled by aspect ratio, it is enforced to the input and output sizes expected by the workflow. It would be better to support a dynamic range of sizes for better aspect ratio. Resolutions must stay within 64 pixel increments and within min/max sizes of the precompiled engine


runner/app/live/api/api.py
```diff
@@ -168,8 +192,60 @@ async def handle_get_status(request: web.Request):
     return web.json_response(status.model_dump())
 
 
+async def handle_restart_pipeline(request: web.Request):
```
eliteprox, line 195: For testing only, can be removed

runner/app/live/streamer/process_guardian.py
```diff
@@ -82,13 +87,33 @@ async def reset_stream(
     ):
         if not self.process:
             raise RuntimeError("Process not running")
+
+        # Check if resolution has changed
```
eliteprox, line 91: Will remove, comfyui pipeline will stop itself

PR #595, opened 2025-05-22
update audio-to-text pipeline and add faster-whisper backend

ad-astra-video:
This PR updates the audio-to-text backend to revert adding flash attention.  Flash Attention is not compatible with word level time stamps and we do not have a way to route audio-to-text requests to workers that are not running Flash Attention.

Updates:
- attention implementation updated to `eager`. `sdpa` attention crashed my runner.
- updates the route to `def` from `async def` to unblock health checks that if failed 3 times will cause the ai-worker to restart the container.
- added [faster-whisper](https://github.com/SYSTRAN/faster-whisper) backend from request from 10xCrazyHorse and Navigarre in streamplace discord
  - I found faster-whisper to be very stable and use less memory compare to transformers.
  - models do not work across the implementations because faster-whisper uses `ctranslate2` but the main models are available for faster-whisper
  - performance timings on 3090ti using 206 second MP4
    ```
       transcription (whisper-large-v3):
       faster-whisper  15.28s   Systran/faster-whisper-large-v3
       transformers    7.68s

       transcription (whisper-large-v3-turbo via deepdml/faster-whisper-large-v3-turbo-ct2 and deepdml/whisper-large-v3-turbo):
       faster-whisper  2.61s
       transformers    4.55s

       word level transcription (whisper-large-v3-turbo):
       faster-whisper  3.51s
       transformers    4.85s
  
       word level transcription (whisper-large-v3):
       faster-whisper  18.00s
       transformers     7.84s
    ```

### DIFF
--- a/runner/app/pipelines/audio_to_text.py
+++ b/runner/app/pipelines/audio_to_text.py
@@ -65,97 +65,117 @@ class AudioToTextPipeline(Pipeline):
 
         torch_device = get_torch_device()
 
-        # Enable FlashAttention based on device compatibility.
-        attn_implementation = "eager"
-        if torch_device.type == "cuda":
-            device = torch.cuda.get_device_properties(0)
-            major, minor = device.major, device.minor
-            # FlashAttention requires CUDA Compute Capability >= 8.0.
-            if (major, minor) >= (8, 0):
-                attn_implementation = "flash_attention_2"
-            else:
-                attn_implementation = "sdpa"
-                logger.warning(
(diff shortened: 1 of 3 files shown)


JJassonn69:
Came across a bug when using .`wav` files in lmps, created a issue #https://github.com/livepeer/lpms/issues/435. Everything else seem to work fine in my testing. LGTM

JJassonn69 (approved):
LGTM

PR #598, opened 2025-05-22
streamer: Avoid numpy back-and-forth to crop image

victorges:
This was spending a.lmost 10ms (my dev server) for every frame.

Improved it as an attempt to improve FPS, but turns out that the multiprocessing
architecture is really efficient and it only improves FPS when the bottleneck is the
`infer.py` process itself (noop pipeline) but not when it is the pipeline process (comfy).

Still an improvement anyway so I'm keeping. Won't mind optimizing the image resize
tho which also takes a lot of time but won't be used with trickle anyway, since the decoder
already reformats to Wx512xH

### DIFF
--- a/runner/app/live/streamer/streamer.py
+++ b/runner/app/live/streamer/streamer.py
@@ -3,7 +3,7 @@ import logging
 import os
 import time
 import numpy as np
-from typing import AsyncGenerator, Awaitable
+from typing import AsyncGenerator, Awaitable, cast
 from asyncio import Lock
 
 import cv2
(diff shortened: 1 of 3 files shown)


runner/app/live/streamer/streamer.py
```diff
@@ -165,22 +165,18 @@ async def run_ingress_loop(self):
...
-
                 # Crop to the center square if image not already square
                 square_size = min(width, height)
                 if width != height:
                     start_x = width // 2 - square_size // 2
                     start_y = height // 2 - square_size // 2
-                    frame_array = frame_array[
-                        start_y : start_y + square_size, start_x : start_x + square_size
-                    ]
+                    frame = frame.crop((start_x, start_y, start_x + square_size, start_y + square_size))
```
j0sh, line 173: This doesn't require a numpy conversion?

j0sh (approved):
Just checking my understanding: with https://github.com/livepeer/ai-runner/pull/564 we set the smaller dimension to 512px, and preserve the aspect ratio along the larger dimension. With this change, we should not have to convert to numpy after cropping the larger dimension to 512.

BTW the swscale resizer (which is what the decoder uses) is plenty fast and doesn't require any conversion, if we wanted to avoid that step to squeeze out more performance. IIRC swscale is what the C++ opencv code used (or at least, it did in the past).

At some point though, I should just look at doing single-pass cropping + resizing after decoding, without needing an additional step here.

PR #599, opened 2025-05-22
Vg/feat/low copy

victorges:


runner/app/live/pipelines/comfyui.py
```diff
@@ -73,8 +71,11 @@ async def initialize(self, **params):
         logging.info("Pipeline initialization and warmup complete")
 
     async def put_video_frame(self, frame: VideoFrame, request_id: str):
-        image_np = np.array(frame.image.convert("RGB")).astype(np.float32) / 255.0
-        frame.side_data.input = torch.tensor(image_np).unsqueeze(0)
+        tensor = frame.tensor
+        if tensor.is_cuda:
+            # Clone the tensor to be able to send it on comfystream internal queue
+            tensor = tensor.clone()
```
varshith15, line 77: is cloning needed?

runner/app/live/streamer/process.py
```diff
@@ -226,9 +226,11 @@ async def _input_loop(self, pipeline: Pipeline):
     async def _output_loop(self, pipeline: Pipeline):
         while not self.is_done():
             try:
-                output_frame = await pipeline.get_processed_video_frame()
-                output_frame.log_timestamps["post_process_frame"] = time.time()
-                self._queue_put_fifo(self.output_queue, output_frame)
+                output = await pipeline.get_processed_video_frame()
+                if isinstance(output, VideoOutput) and not output.tensor.is_cuda:
+                    output = output.replace_tensor(output.tensor.cuda())
```
varshith15, line 231: we don't need it on gpu for output right?

runner/app/live/trickle/encoder.py
```diff
@@ -91,7 +91,12 @@ def custom_io_open(url: str, flags: int, options: dict):
                 continue
             avframe.log_timestamps["frame_end"] = time.time()
             log_frame_timestamps("Video", avframe.frame)
-            frame = av.video.frame.VideoFrame.from_image(avframe.image)
+
+            tensor = avframe.tensor.squeeze(0)
+            image_np = (tensor * 255).byte().cpu().numpy()
```
varshith15, line 96: this is just additional .cpu(), if you don't do .cuda() above?

varshith15 (approved):
LGTM! just that additional .cuda() i don't fully understand

runner/app/live/trickle/encoder.py
victorges: I'm not sure, I honestly didn't try changing this logic a lot, only copied from the comfyui pipeline. I didn't bother optimizing this too much since this already runs on a separate thread in the outer process, which is not the bottleneck rn, so it should be fine as is

runner/app/live/streamer/process.py
victorges: Just re-ran some tests for comparison:
- Keeping this `.cuda()` we get ~21FPS
- Changing to `share_memory_()` (shared CPU mem) we get ~20FPS
- Removing this altogether we also get 17 FPS

I did a bunch of further testing cause it seems counter-intuitive, but the gist is:
- `.cuda()` is blazingly fast, taking ~0.4ms on average and optimizing the tensor transfer over the queue
- `share_memory()` can be pretty slow, up to 7ms if we call it on every tensor
   - But there's some weird stuff going on since the tensors are coming from comfystream on the same address as previous tensors. If we add some dirty logic to avoid calling `share_memory_` when it's already shared it could go up to 20 FPS, which is still slower than `.cuda()` anyway.

That said, I'll keep the `cuda()` for now which seems to be worth it. One caveat is that this might be different per machines as well, so at some point ideally we could have some metrics in production for all of this.

PR #616, opened 2025-05-30
pipelines: Bring back the native streamdiffusion pipeline

victorges:
This is to bring back the `streamdiffusion` pipeline that had been removed on #461.

The process was mainly to revert the relevant changes and then adjust any interface that changed
so it works on the new world. We have actually made several improvements to the pipeline since then
and this works much more performant than before.

**On my dev server, the comfystream pipeline performs 21 FPS while this works at 40 FPS.**

Use the box to test this: https://github.com/livepeer/go-livepeer/pull/3602


victorges (commented):
LGTM! Thanks for this

PR #632, opened 2025-06-05
update openapi spec for new version

ad-astra-video:
Update openapi spec auto gen files for version bump

cc @leszko


eliteprox (approved):
lgtm

PR #649, opened 2025-06-23
dl_checkpoints: Create PIPELINE option for dev

victorges:
This creates an option to set a PIPELINE env to only download/build the models for a specific live pipeline.

This is useful on local development so you can iterate quickly when making changes to a specific pipeline models. Follows same env used by the box itself.


victorges:
I'm insta-merging cause I need this feature for a guide for external devs to write pipelines. Happy to address any post-merge comments!

PR #652, opened 2025-06-26
streamdiffusion: Use wrapper with controlnet support

victorges:
This migrates from using a copy of the StreamDiffusionWrapper to using one exported by the
library directly, so we don't need to re-implement anything on our side. This support was added
on https://github.com/livepeer/StreamDiffusion/pull/3 into the controlnet version of the lib.


ryanontheinside (approved):
lgtm

PR #666, opened 2025-07-04
process: Make sure params don't accumulate in queue

victorges:
This is to add support to a high frequency stream of updates which might start
happening with the TD plugin being developed.

The idea is to make sure we always update to the latest provided params, instead
of going through each of the provided params step by step. So if the param updates
are not being processed fast enough, we should just skip to the latest instead of
letting them accumulate on the queue.

Also added some timing metric to have some idea how much these are taking.

### DIFF
--- a/runner/app/live/streamer/process.py
+++ b/runner/app/live/streamer/process.py
@@ -158,7 +158,7 @@ class PipelineProcess:
     async def _initialize_pipeline(self):
         try:
             params = await self._get_latest_params(timeout=0.005)
-            if params:
+            if params is not None:
                 logging.info(f"PipelineProcess: Got params from param_update_queue {params}")
             else:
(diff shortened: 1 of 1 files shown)


eliteprox (approved):
LGTM! Param updates are working and consistent

runner/app/live/streamer/process.py
```diff
@@ -238,17 +238,44 @@ async def _output_loop(self, pipeline: Pipeline):
     async def _param_update_loop(self, pipeline: Pipeline):
         while not self.is_done():
             try:
-                params = await asyncio.to_thread(self.param_update_queue.get, timeout=0.1)
+                params = await self._get_latest_params(timeout=0.1)
+                if not params:
```
victorges, line 242: Need to check is None instead

PR #668, opened 2025-07-08
streamdiffusion: Update controlnets scale dynammically

victorges:
These are handled separately from the other stream params so let's implement it too


ryanontheinside (approved):
lgtm

PR #670, opened 2025-07-09
live: Log publish duration and some trickle logging cleanup

j0sh:
We're seeing a lot of slow subscribes, so add some logging to determine whether the slowness is on the publish end.


j0sh:
I'm going to sneak a slight behavioral change in now - cc @victorges :

[Set trickle publisher output queue size to 1.](https://github.com/livepeer/ai-runner/pull/670/commits/0d036b627cd19badf2a3de416fe9a651d00247d7) 
 
Previously, if there was a networking issue between the runner and
orchestrator, we would queue up data indefinitely with no feedback.
Rather than do this, set the queue size to 1 and create some
back-pressure which should be reflected in the timing / logging
statistics.

PR #673, opened 2025-07-09
Update Dockerfiles for StreamDiffusion pipeline and dependencies

hjpotter92:


runner/docker/Dockerfile.live-base
```diff
@@ -1,32 +1,29 @@
 # Based on https://github.com/huggingface/api-inference-community/blob/main/docker_images/diffusers/Dockerfile
-FROM nvidia/cuda:12.6.3-cudnn-devel-ubuntu22.04
+FROM nvidia/cuda:12.8.1-cudnn-devel-ubuntu22.04
```
victorges, line 2: Btw, unrelated to this change, I'd love to at some point optimizing these images by using the runtime image instead of devel. I think we only really need `devel` to build ffmpeg below, so we could do a multi-stage and reduce our runtime image size by at least 2GB

victorges (approved):
lgtm

runner/docker/Dockerfile.live-app__PIPELINE__
```diff
@@ -1,28 +1,42 @@
...
-RUN apt-get update && apt-get install -y \
+RUN apt update && apt install -yqq \
     wget \
     libcairo2-dev \
     libgirepository1.0-dev \
     pkg-config \
-    && apt-get clean && rm -rf /var/lib/apt/lists/*
+    && apt clean && rm -rf /var/lib/apt/lists/*
+
+RUN --mount=type=bind,source=requirements.live-ai.txt,target=/tmp/requirements.txt \
```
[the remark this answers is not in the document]
hjpotter92: > Bind mounts are more efficient than COPY for including files from the build context in the container. Note that bind-mounted files are only added temporarily for a single RUN instruction, and don't persist in the final image. If you need to include files from the build context in the final image, use COPY.

I believe we don't need the requirements file in final docker image; so used the one-off mount and pip install.

runner/docker/Dockerfile.live-app__PIPELINE__
```diff
@@ -1,28 +1,42 @@
...
+    conda run -n comfystream pip install --root-user-action ignore --requirement /tmp/requirements.txt
+
+RUN conda run -n comfystream pip install --no-cache-dir --force-reinstall \
+    torch \
+    torchvision \
+    torchaudio \
+    pytorch_triton \
```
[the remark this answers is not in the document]
hjpotter92: yeah, i was originally planning to create a conda environment file (`yaml`) with all these pinned and predefined; but am unable to test it on my teapot of a system and couldn't keep pushing commits to github just to get docker builds :|

dropped the pinned version from earlier commit from @JJassonn69 (which was pinning to march's nightly builds) as the discussion on pytorch forums mentioned that support for blackwell landed couple weeks back in stable releases.

.github/workflows/ai-runner-docker.yaml
```diff
@@ -28,6 +28,12 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
```
pwilczynskiclearcode, line 35: What about `push:` and `workflow_dispatch:` events?

runner/Dockerfile
```diff
@@ -1,45 +1,46 @@
...
+RUN source /root/.bashrc && \
+    pyenv install $PYTHON_VERSION && \
+    pyenv global $PYTHON_VERSION && \
+    pyenv rehash
 
 # Upgrade pip and install your desired packages
-ARG PIP_VERSION=23.3.2
 RUN pip install --no-cache-dir --upgrade pip==${PIP_VERSION} setuptools==69.5.1 wheel==0.43.0 && \
-  pip install --no-cache-dir torch==2.1.1 torchvision==0.16.1 torchaudio==2.1.1
+    pip install --no-cache-dir torch==2.1.1 torchvision==0.16.1 torchaudio==2.1.1
```
pwilczynskiclearcode, line 35: why don't we place these dependencies on requiremens.txt?

pwilczynskiclearcode (approved):

runner/Dockerfile
hjpotter92: tried it locally. quite difficult to do that without having proper working dev setup. there are too many conflicts
```

#11 654.3 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
#11 654.3 gradio 5.33.1 requires fastapi<1.0,>=0.115.2, but you have fastapi 0.111.0 which is incompatible.
#11 654.3 gradio 5.33.1 requires python-multipart>=0.0.18, but you have python-multipart 0.0.9 which is incompatible.
#11 654.3 gradio 5.33.1 requires starlette<1.0,>=0.40.0; sys_platform != "emscripten", but you have starlette 0.37.2 which is incompatible.
#11 654.3 mediapipe 0.10.21 requires numpy<2, but you have numpy 2.1.2 which is incompatible.
#11 654.3 xformers 0.0.30 requires torch==2.7.0, but you have torch 2.7.1+cu128 which is incompatible.
```

hjpotter92:
merging this PR.

for addressing the conda env and better requirements pinning; i would suggest moving to `uv` based management or if we want to keep conda for development purposes, bunding its own `environment.yaml` file alongside.

i tried updating the requirements file and build locally, but that results in a few broken dependency issues: https://github.com/livepeer/ai-runner/pull/673#discussion_r2199693672

cc @victorges @pwilczynskiclearcode

PR #700, opened 2025-07-24
streamdiffusion: Add some explicit validation for t_index_list

victorges:
We were getting some cryptic behavior without this. Just like this description.

Fixes INF-275

### DIFF
--- a/runner/app/live/pipelines/streamdiffusion.py
+++ b/runner/app/live/pipelines/streamdiffusion.py
@@ -132,11 +132,18 @@ class StreamDiffusionParams(BaseModel):
     def check_t_index_list(model: "StreamDiffusionParams") -> "StreamDiffusionParams":
         if not (1 <= len(model.t_index_list) <= 4):
             raise ValueError("t_index_list must have between 1 and 4 elements")
-        for idx in model.t_index_list:
-            if not (0 <= idx <= model.num_inference_steps):
+
+        for value in model.t_index_list:
+            if not (0 <= value <= model.num_inference_steps):
                 raise ValueError(
-                    f"Each t_index_list value must be between 0 and num_inference_steps ({model.num_inference_steps})"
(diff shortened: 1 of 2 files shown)


runner/app/live/pipelines/streamdiffusion.py
```diff
@@ -127,6 +127,18 @@ class Config:
...
 
+    @model_validator(mode="after")
+    @staticmethod
+    def check_t_index_list(model: "StreamDiffusionParams") -> "StreamDiffusionParams":
+        if not (1 <= len(model.t_index_list) <= 4):
+            raise ValueError("t_index_list must have between 1 and 4 elements")
+        for idx in model.t_index_list:
+            if not (0 <= idx <= model.num_inference_steps):
+                raise ValueError(
+                    f"Each t_index_list value must be between 0 and num_inference_steps ({model.num_inference_steps})"
```
gioelecerati, line 138: Lovely!
A question, what are the dimensions of the t_index_list? let's say, what is the difference between [0,10,20] and [10,10,10]? Right now we're updating the values linearly, treating them as a single value, but I bet it would be better to have them separate?

gioelecerati (approved):
LGTM

runner/app/live/pipelines/streamdiffusion.py
victorges: The values are the value sent to the diffusion model during each inference step. Users need to have complete control of each values in `t_index_list` as they have the most dramatic effect on the output, like being more adherent to the prompt or input image, creativity etc. I sent some details on discord [here](https://discord.com/channels/423160867534929930/1397471786865655859/1397966375472988282).

I see you've already implemented that https://github.com/livepeer/pipelines/pull/1043 so great :D

victorges: Oh one other thing to note is that the values must be in ascending order. Let me add this validation here as well.

PR #716, opened 2025-08-04
streamdiffusion: Add SD1.5 support

victorges:
Using `PerfectPhotonV2.1` for now but we can add more models easily.

### DIFF
--- a/runner/app/live/pipelines/streamdiffusion_params.py
+++ b/runner/app/live/pipelines/streamdiffusion_params.py
@@ -66,7 +66,7 @@ class StreamDiffusionParams(BaseModel):
     model_id: Literal[
         "stabilityai/sd-turbo",
         "varb15/PerfectPhotonV2.1",
-    ] = "varb15/PerfectPhotonV2.1"
+    ] = "stabilityai/sd-turbo"
     """Base U-Net model to use for generation."""
 
     # Generation parameters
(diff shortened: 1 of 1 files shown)


runner/dl_checkpoints.sh
```diff
@@ -129,15 +129,18 @@ function download_streamdiffusion_live_models() {
...
+  huggingface-cli download varb15/PerfectPhotonV2.1 --include "*.safetensors" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
 
-  # ControlNet models
+  # SD2.1 (turbo) ControlNet models
   huggingface-cli download thibaud/controlnet-sd21-openpose-diffusers --include "*.bin" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
   huggingface-cli download thibaud/controlnet-sd21-hed-diffusers --include "*.bin" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
   huggingface-cli download thibaud/controlnet-sd21-canny-diffusers --include "*.bin" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
   huggingface-cli download thibaud/controlnet-sd21-depth-diffusers --include "*.bin" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
   huggingface-cli download thibaud/controlnet-sd21-color-diffusers --include "*.bin" "*.json" "*.txt" --exclude ".onnx" ".onnx_data" --cache-dir models
+  # SD1.5 controlnet models
```
pwilczynskiclearcode, line 141: should we have it as a separate build like `PIPELINE=streamdiffusion-1.5`?

pwilczynskiclearcode: Ah ok. I thought it's a kind of fork

pwilczynskiclearcode (approved):

runner/docker/Dockerfile.live-base-streamdiffusion
```diff
@@ -52,7 +52,7 @@ WORKDIR /app
 
 # Disable HF hub connectivity to avoid internet access during runtime.
 # This is manually disabled on the script to download models (dl_checkpoints.sh)
-ENV HF_HUB_OFFLINE=1
+ENV HF_HUB_OFFLINE=0
```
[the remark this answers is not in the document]
victorges: I'm just trying to do that, but the model was breaking on the offline mode as it complains some file is missing (but it just doesn't exist). I'll give it a try but worst case will just merge as is and take the tech debt (I'll add a TODO here so we can undo this)

runner/app/live/streamer/process.py
```diff
@@ -170,6 +170,7 @@ async def _initialize_pipeline(self):
                 return pipeline
         except Exception as e:
             self._report_error(f"Error loading pipeline: {e}")
+            logging.exception(e)
```
j0sh, line 173: tiny nit: passing in the exception is not needed here, since `logging.exception` already prints out the exception that's in-scope 🪄 

something like `logging.exception("Error loading pipeline")` should be perfectly fine.

PR #718, opened 2025-08-05
feat: adds logs to check streamdiffusion load times

varshith15:


victorges:
The loading times are also extremely consistent between dev environment and staging/production (which are just TensorDock) anyway. If you set up a proper dev env with the box, you can add all the debug logs you want to debug it.

PR #726, opened 2025-08-08
streamdiffusion: Set HF to offline mode in runtime

victorges:
- Offline mode by default on the image
- On the dl_checkpoints script we disable it just in case any file is missing during compilation (most are downloaded by the script itself, but let's not be obnoxious)


varshith15 (approved):
LGTM

PR #730, opened 2025-08-13
feat: dynamic t_index_list

varshith15:
Adds support to dynamically update t_index_list instead of switching engines

### DIFF
--- a/runner/docker/Dockerfile.live-base-streamdiffusion
+++ b/runner/docker/Dockerfile.live-base-streamdiffusion
@@ -29,7 +29,7 @@ RUN conda run -n comfystream pip install --no-cache-dir --force-reinstall \
     conda run -n comfystream pip install --no-cache-dir \
     xformers==0.0.30 --no-deps
 
-# Install StreamDiffusion @ v0.0.1-cnet.7 into the comfystream environment
+# Install StreamDiffusion lib into the comfystream environment
 RUN conda run -n comfystream pip install git+https://github.com/livepeer/StreamDiffusion.git@v0.0.1-cnet.8#egg=streamdiffusion[tensorrt]
 
 # Pin versions of ONNX runtime which are too loose on streamdiffusion setup.py


runner/app/live/pipelines/streamdiffusion.py
```diff
@@ -50,6 +50,8 @@ class Config:
     delta: float = 0.7
     num_inference_steps: int = 50
     t_index_list: List[int] = [12, 20, 32]
+    min_batch_size: int = 1
+    max_batch_size: int = 4
```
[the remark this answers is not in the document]
varshith15: i think its still better that we keep it as a configurable thing without it being exposed to the end users, just removing from the streamdiffusion params should do, ill just pass them just like the engine_dir

WDYT?

runner/app/live/pipelines/streamdiffusion.py
```diff
@@ -343,15 +341,15 @@ def _prepare_controlnet_configs(params: StreamDiffusionParams) -> Optional[List[
     return controlnet_configs
 
 
-def load_streamdiffusion_sync(params: StreamDiffusionParams, engine_dir = "engines", build_engines_if_missing = False):
+def load_streamdiffusion_sync(params: StreamDiffusionParams, min_batch_size: int, max_batch_size: int, engine_dir = "engines", build_engines_if_missing = False):
```
victorges, line 344: I think we need the default here as well since we don't send it on this pipeline
```suggestion
def load_streamdiffusion_sync(params: StreamDiffusionParams, min_batch_size = 1, max_batch_size = 4, engine_dir = "engines", build_engines_if_missing = False):
```

runner/docker/Dockerfile.live-base-streamdiffusion
```diff
@@ -30,7 +30,7 @@ RUN conda run -n comfystream pip install --no-cache-dir --force-reinstall \
     xformers==0.0.30 --no-deps
 
 # Install StreamDiffusion @ v0.0.1-cnet.7 into the comfystream environment
```
victorges, line 32: Gotta update the comment too. Let me just remove that it's so annoying and redundant anyway, probably written by AI and I approved it lol
```suggestion
# Install StreamDiffusion lib into the comfystream environment
```

victorges (approved):
LGTM!

PR #754, opened 2025-08-25
process: Ensure stream is stopped with no inputs

victorges:
It might happen that we never get any input from the stream, we might be forever stuck in the
`DEGRADED_INPUT` state because of [this check](https://github.com/livepeer/ai-worker/blob/c3396d13f00d7d7bb4f831d3503ee0da44fb2005/runner/app/live/streamer/process_guardian.py#L199-L200) that comes before any timeout.

To fix it, I made sure we check for the input timeout before checking for the pipeline loading
conditions. The only side-effect I had to include was making the `time_since_last_input` start
as `0` when the pipeline starts, instead of "+INF" (`current_time`). I've corrected for this
on the only place that I think it'd matter, in the stream shutdown branch.

Implements ENG-3004

### DIFF
--- a/runner/app/live/streamer/process_guardian.py
+++ b/runner/app/live/streamer/process_guardian.py
@@ -172,11 +172,11 @@ class ProcessGuardian:
                 logging.info(
                     f"Shutting down streamer. Flagging DEGRADED_INPUT state during shutdown: time_since_last_input={time_since_last_input:.1f}s"
                 )
-            return (
-                PipelineState.DEGRADED_INPUT
-                if time_since_last_input < 90
-                else PipelineState.ERROR  # Not shutting down after 30s, declare ERROR
-            )
+            if time_since_last_input < 90:
+                return PipelineState.DEGRADED_INPUT
+            # Stream hasn't shut down 30s after the trigger above (90s total idle time).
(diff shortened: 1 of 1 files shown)


runner/app/live/streamer/process_guardian.py
```diff
@@ -158,18 +158,28 @@ def _compute_current_state(self) -> str:
...
+                logging.info(
+                    f"Shutting down streamer. Flagging DEGRADED_INPUT state during shutdown: time_since_last_input={time_since_last_input:.1f}s"
+                )
             return (
-                PipelineState.OFFLINE
-                if time_since_last_input > 3  # 3s grace period after shutdown
-                else PipelineState.DEGRADED_INPUT
+                PipelineState.DEGRADED_INPUT
+                if time_since_last_input < 90
+                else PipelineState.ERROR  # Not shutting down after 30s, declare ERROR
```
j0sh, line 178: should this say 90s instead of 30s?

j0sh (approved):
LGTM

runner/app/live/streamer/process_guardian.py
victorges: it was referring to the 30s after the trigger above, so 90s total. I elaborated the comment further so it's clearer

PR #760, opened 2025-08-30
streamdiffusion: Add SDXL support

victorges:


runner/app/live/pipelines/streamdiffusion.py
```diff
@@ -284,12 +279,38 @@ def _prepare_controlnet_configs(params: StreamDiffusionParams) -> Optional[List[
...
+    dir = 'sdxl_models' if is_sdxl else 'models'
+
+    if not ip_cfg.ipadapter_model_path:
+        match ip_cfg.type:
+            case 'regular':
+                ip_cfg.ipadapter_model_path = f"h94/IP-Adapter/{dir}/ip-adapter_{arch}.bin" # type: ignore
+            case 'faceid':
+                ip_cfg.ipadapter_model_path = f"h94/IP-Adapter-FaceID/ip-adapter-faceid_{arch}.bin" # type: ignore
+    if not ip_cfg.image_encoder_path:
+        ip_cfg.image_encoder_path = f"h94/IP-Adapter/{dir}/image_encoder" # type: ignore
```
victorges, line 304: FTR had this whole logic implemented for custom ipadapter models for SDXL. Turns out SDXL+IPAdapt is not working on the base lib for now, but I'll keep the logic here as we should be fixing it soon.

PR #775, opened 2025-09-08
Update error message for missing engine file

ad-astra-video:
Update to error to indicate missing engine. 

Ran into this on new machine that had volume mount wrong and ValueError indicated bad parameter but traced to engine missing.


victorges (approved):
LGTM!

PR #776, opened 2025-09-09
streamdiffusion/sd15: Cleanup models and make OpenJourney default

victorges:
After a bunch of testing, decided on going with OpenJourney as the default SD15 model.
It is both less NSFW-leaning as well as more consistent than PerfectPhoton on the tests I made.
Tried a bunch of other models as well (the ones being removed) and it seemed to perform generally best.

### DIFF
--- a/runner/app/live/pipelines/streamdiffusion_params.py
+++ b/runner/app/live/pipelines/streamdiffusion_params.py
@@ -30,16 +30,10 @@ CONTROLNETS_BY_TYPE: Dict[ModelType, List[str]] = {
 
 MODEL_ID_TO_TYPE: Dict[str, ModelType] = {
     "stabilityai/sd-turbo": "sd21",
-    "SG161222/Realistic_Vision_V6.0_B1_noVAE": "sd15",
-    "varb15/PerfectPhotonV2.1": "sd15",
     "stabilityai/sdxl-turbo": "sdxl",
-    "Lykon/dreamshaper-8": "sd15",
     "prompthero/openjourney-v4": "sd15",
-    "dreamlike-art/dreamlike-diffusion-1.0": "sd15",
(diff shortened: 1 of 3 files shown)


victorges:
Some notes when testing the models with different inputs. Had different prompts for each but still tried tuning some params like t-list, controlnet weights etc to look for the best output for each.
---
# Model tests

## prompthero/openjourney-v4

- Dance: Some nice looking output but remembers dreamshaper in a sense.
- Headshot: More consistent output, looks good
- Snakes (abstract): Good results, trippy, colorful, consistent

## Lykon/dreamshaper-8(-lcm)

- Dance: looks good
- Headshot: Looks good too. A little flickery tho
- Snakes: Good stuff. Different style from openjourney which is a bit more realistic.

LCM? No. Same result as regular+LCM, so it's simpler to keep same logic for all SD15 models.

## danbrown/Lyriel-v1-5

- Dance: realistic output, but also imaginative with lower t_index_list
- Headshot: similar to openjourney but more flickery
- Snakes: Not as good, tries to be realistic and ends up washed off

## SG161222/Realistic_Vision_V6.0_B1_noVAE

- Dance: Generates some real washed off output. Eliminated.
- Headshot: Looks ok. Also feels a bit washed off. Kept taking my shirt off.
- Snakes: Actually trippy but not very consistent. Sry you lost by now realvis

## varb15/PerfectPhotonV2.1

Pervert.

## dreamlike-art/dreamlike-diffusion-1.0

Proprietary.

- Dance: Nothing special

## dreamlike-art/dreamlike-photoreal-2.0

Proprietary.

- Dance: Nothing special

## stablediffusionapi/deliberate-v2

Proprietary. CC BY-NC-ND 4.0

- Dance: same as Lyriel.

PR #796, opened 2025-09-15
streamdiffusion: Report internal errors updating parameters

victorges:
This makes sure we propagate internal errors when updating the parameters so that
feedback can be shown to the user. Made it through a special logging field to be added
when we want the error to go to the end user, like for the IPAdapter errors we are getting
in the library. This should not be abused as to become pure spam tho, only actionable/critical
or high-level errors should be reported to the user.


ryanontheinside (approved):
lgtm

PR #809, opened 2025-09-24
infer.py: Avoid leaving orphaned process behind

victorges:
This is to avoid having the pipeline process resist a restart when the parent process dies for any reason.

Fixed this in 2 ways:
- Ensure we gracefully stop the process guardian when we have an uncaught exception (cause of orphaned process recently)
- Make pipeline process kill itself when it notices the parent process is gone. This is implemented with a unix signal and a wathdog thread

Fixes INF-350


victorges (commented):
self-review ✅

PR #896, opened 2025-12-29
routes: Fix version and hardware metadata routes

victorges:
When running a custom pipeline, one still needed to set the `PIPELINE` and `MODEL_ID`
env vars which is super annoying, especially when running locally with `uv run`.

This avoids that by returning the `PIPELINE` and `MODEL_ID` configurations from the
actual running pipeline, in case the envs are not set. We should consider even deprecating
the envs altogether, but I'm trying to make smaller changes since I won't be around to test
this further.


mjh1:
✅

PR #900, opened 2026-03-07
feat(live): add @pipeline decorator for simplified pipeline creation

rickstaa:
## Summary

Add a `@pipeline` decorator that turns a plain function or class into a fully working live `Pipeline` subclass — handling frame queues, lifecycle management, parameter validation, and thread safety automatically.

### Function form (minimal)

```python
from runner.app import start_app
from runner.live.pipelines import pipeline, BaseParams
from runner.live.trickle import VideoFrame

@pipeline(name="green-shift")
async def green_shift(frame: VideoFrame, params: BaseParams) -> torch.Tensor:
    tensor = frame.tensor.clone()
    tensor[:, :, :, 1] = torch.clamp(tensor[:, :, :, 1] + 0.3, -1.0, 1.0)
    return tensor

if __name__ == "__main__":
    start_app(pipeline=green_shift._spec)
```

### Class form (with lifecycle hooks)

```python
@pipeline(name="edge-detect", params=EdgeParams)
class EdgeDetect:
    def on_ready(self, **params):
        """Called once at startup — load models, allocate GPU resources."""
        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
        self.kernel = build_sobel_kernel(self.device)

    def transform(self, frame: VideoFrame, params: EdgeParams) -> torch.Tensor:
        """Called per frame — run inference."""
        return sobel_edge_detect(frame.tensor, self.kernel, params.threshold)

    def on_update(self, **params):
        """Called when params change mid-stream."""
        self._threshold = params.get("threshold", 0.1)

    def on_stop(self):
        """Called on shutdown — cleanup."""
        pass
```

### What the decorator handles

- Wraps functions/classes into a `Pipeline` subclass with async frame queues
- Both sync and async `transform` methods (sync runs in thread pool)
- Auto-generates `PipelineSpec` for dynamic loading via `importlib`
- Thread-safe parameter updates via `asyncio.Lock`
- `prepare_models()` classmethod for build-time model downloads
- Warns if `@pipeline` decorates a nested definition (dotted `__qualname__` breaks the loader)

### Includes two example pipelines

- **green_shift**: function form — boosts the green channel (no GPU needed)
- **edge_detect**: class form — Sobel edge detection with all lifecycle hooks

### How to test it works

1. See description

You can test both examples and will get a result like

image

image


docs/custom-pipeline.md
```diff
@@ -74,63 +74,177 @@ touch src/my_pipeline/pipeline/params.py
...
+class DepthMidas:
+
+    def on_ready(self, **params):
+        # Load model and set up device
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        self.model = torch.hub.load("intel-isl/MiDaS", "MiDaS_small", pretrained=True)
+        self.model.to(self.device).eval()
+        self.transforms = torch.hub.load("intel-isl/MiDaS", "transforms").small_transform
+
+    def transform(self, frame: VideoFrame, params: DepthParams) -> torch.Tensor:
```
victorges: Shouldn't this be async as well just in case, to support async workflows? Shouldn't be a problem if the implementation is not async and just returns the result, but more flexible

docs/custom-pipeline.md
```diff
@@ -405,6 +519,19 @@ The orchestrator will route requests to your local runner at `http://localhost:8
 
 ---
 
+## Publishing to the Marketplace
+
+Publish your pipeline to make it discoverable on the marketplace. The CLI extracts the parameter schema from your `@pipeline` decorator and registers it automatically.
+
+```bash
+livepeer publish examples/pipelines/depth_midas.py
```
victorges: Wowwww this seems like the coolest part and it's just a detail on this PR?

runner/src/runner/live/pipelines/create.py
```diff
@@ -0,0 +1,134 @@
...
+            f"@pipeline class {user_cls.__name__} must define a 'transform' method"
+        )
+
+    has_on_ready = hasattr(user_cls, "on_ready")
+    has_on_update = hasattr(user_cls, "on_update")
+    has_on_stop = hasattr(user_cls, "on_stop")
+    has_prepare = hasattr(user_cls, "prepare_models")
+    label = user_cls.__name__
+
+    class GeneratedPipeline(Pipeline):
```
victorges, line 107: You know the cleanest approach would be to simply migrate to existing Pipeline base class and all the implementations to the new interface. I don't think there's a core need for spring 2 class interfaces and we could probably stick to just the cleaner one. If there are more "hardcore" functions that not everyone needs to override, the override can be optional and we just have sane defaults if not implemented. 

So instead of having 1 function and 2 class forms, it's either function or class. With decorator always used for registration. Or alternatively there could be at least a "Simple pipeline" base class so the interface is not just assumed by the decorator but properly defined and typed.

Not required change, just an architectural nit. I'm not even a core maintainer anymore hahaha

victorges (approved):
LGTM!

docs/custom-pipeline.md
rickstaa: Oh yeah, this was part of my follow-up pull request. I’m still iterating on the schema and need to check with Adam first to see what makes the most sense for the UI generation.

runner/src/runner/live/pipelines/create.py
rickstaa: Good point, it's probably better to define an interface and apply it to the Pipeline class. Let’s treat this as a draft for revisiting the idea later.

